PR #9, opened 2015-06-22
Move text to atlas-strings.xml

garhiggins:
This will allow for localization of text.


sjones94549:
would you mind squashing and rebasing agains master?

sjones94549:
Text moved to string resources in `0.2.0`

PR #24, opened 2015-10-21
Change bloating default colors + few more improvements

Enuviel:
Improvements:
- conversations: avatar icons turns from light gray to dark gray and finally to black (clean canvas before draw). nice separator
- messages: white text on blue background instead of black
- composer: left margin for text
- picker: left margin for text and hidden "+" button (doesn't work. cannot find implementation?)


sjones94549:
Thanks @Enuviel -- we've updated Atlas significantly in 0.2.0 and can't merge this.  We're anticipating stable APIs from here on :)

PR #25, opened 2015-11-06
Update sdk version to 15

raphaelcuba:
Update sdk version to minimum be 15.


sjones94549:
We're keeping the minimum Android version at 14

PR #40, opened 2015-12-02
Styles refactoring and other minor changes

bryant1410:
I think the best approach for custom styling is to define styles for (almost?) every widget. This way there are 3 levels of customization:

* Colors: someone can just change some colors
* Widget style: someone can change the style of a view in a layout, for example the style of the time label in a conversation item
* Layout: this can always be changed by overriding the layout with other with the same name

In this PR I set some styles, but obviously not all. This work must be continued.


bryant1410:
I should clean a little bit this PR

bryant1410:
This is getting some commits that have nothing to do with the original PR, so I'm closing this

PR #47, opened 2016-01-27
Async provider

Drac0n:
Participant provider is now asynchronous. This way you don't have to load all participants from database at the start of application.


Drac0n:
I don't recommend merge yet, it seems there is a bug. When adding participants to conversation, some components don't refresh participant list.

sriamar:
Hello @Drac0n 

Thanks for sending this over, and apologies for the very late followup. With the support for identities, we have stopped using Participant. Also, we have removed flavors in latest versions of Atlas. 

We are also working on improving Layer API's to be async in general. You will see those changes in the next few months. 

So, this PR is no longer valid for Atlas. So, I am closing this. Please feel free to reach out with any thoughts / comments to Layer support via https://support.layer.com/.

PR #55, opened 2016-02-19
Display timestamp for first message in the list

jasco:
This change was also requested by issue layerhq/Atlas-Android/#45.

### DIFF
--- a/layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
+++ b/layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
@@ -335,16 +335,10 @@ public class AtlasMessagesAdapter extends RecyclerView.Adapter<AtlasMessagesAdap
         if (cluster.mClusterWithPrevious == null) {
             // No previous message, so no gap
             viewHolder.mClusterSpaceGap.setVisibility(View.GONE);
-            viewHolder.mTimeGroup.setVisibility(View.GONE);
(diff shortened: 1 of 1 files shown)


layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
```diff
@@ -278,16 +278,10 @@ public void bindCellViewHolder(CellViewHolder viewHolder, int position) {
         if (cluster.mClusterWithPrevious == null) {
             // No previous message, so no gap
             viewHolder.mClusterSpaceGap.setVisibility(View.GONE);
```
andela-wakinsanmi: Hello @jasco 

Thank you for the PR, will it be better if move this to the bindDateTimeForMessage.

andela-wakinsanmi:
Can you please Fix the conflicts and change the base to develop? 

Thanks

jasco:
I no longer use this project.  In the spirit of helping out I can see what I can do, but it may take some time.

jasco:
I made the changes requested.  The merge appears okay, but I have not tested the results.

I noticed upstream changed from displaying the sent timestamp to the received timestamp.  This seems like an improvement and the right thing to do for received messages, but I wonder if sent messages should still show sent timestamp.  I have been away from the software for a long time so my contextual understanding is limited.  For instance, I do not know if the received timestamp is eventually populated for sent messages or not.

andela-wakinsanmi:
Thanks @jasco for taking time to fix the merge conflict.

PR #60, opened 2016-03-09
atlas android listener implementaion

BabuinGH:
@sjones94549 : I had issues while updating the participant information via the ParticipantListener as the listener was never been called by the atlas-android.  Hence after doing the below modifications, I am able to update the participant information correctly. Please review.


sjones94549:
Good call; what say you to:

1. Remove the `ParticipantListener` interface from both the `RailsParticipantProvider` and `DemoParticipantProvider`;
2. Add `@Override` to both `registerParticipantListener` and `unregisterParticipantListener` in each.

sriamar:
Hello,

We have changed the implementation of Atlas and removed the different flavors. So, this PR is not valid anymore. If there are any issues in this area, please feel free to reach out to us via https://support.layer.com

Thanks
Amar

PR #62, opened 2016-04-04
Null and empty check for getInitials

rhall1705:


layer-atlas/src/main/java/com/layer/atlas/util/Util.java
```diff
@@ -127,6 +127,7 @@ public static String getInitials(Participant p) {
     }
 
     public static String getInitials(String fullName) {
+        if(fullName == null || fullName.isEmpty()) return "";
```
sriamar: This should happen only when participant provider doesn't set full name. Why does the participant provider not include full name? 

I am tempted to think that this should stay as is and the handling of `null` full-name should be done at Participant provider level. Thoughts?

Is there a valid scenario where full-name can be absent for an user?

nullptr2this: I'm not sure if it's considered a "valid" scenario, but when the name of a user is tracked via conversation metadata the metadata may not have been entirely synced at the point where it is queried from the Layer SDK.  In that case it would be null or empty and once Atlas passes it to this method there is an NPE.

sriamar:
Okay. The impact of merging this are small and it doesn't affect Atlas. So, I am merging this. 

But be aware that if your participant name is empty, this could lead to poor UI state for user.

PR #66, opened 2016-04-22
Reycle sampled bitmap after preview bitmap is compressed

abdyer:
If the sampled bitmap is small enough, the preview bitmap will reference the original rather than creating a new scaled bitmap. This results in a "can't compress a recycled bitmap" error.


sriamar:
Andy, Thanks for the PR. Merged.

PR #69, opened 2016-04-28
Added ability to set colors of bubbles, text, date, underline, & cursor programmatically, & attachment menu has white background with dropshadow

lexi-sr:
…rView. 

Previously, we could only set these colors with the XML file, but in my app, the colors need to be dynamically set at runtime (the color might vary). The other way to do this from the existing app would have been to pass in an AttributeSet to AtlasMessagesRecyclerView.parseStyle, but I couldn't find a good way to generate an AttributeSet instance, so I think this is the best solution. 

Also in my app, the attachment menu was transparent (and not white with the dropshadow like in the Atlas Messenger sample app), so I added a fix to that too.


smpete:
Hi,

If you know the colors before the view has to be instantiated, you can create the view programmatically using a ContextThemeWrapper object and pass that in as the context when creating the RecyclerViews. You'd create the ContextThemeWrapper with a new theme with the appropriate colors. Unfortunately, Android does not let you change the theme/style of a view after it has been instantiated. If you need to change the colors dynamically after the view is instantiated then we would need to expose setters for most if not all of the colors in that view.

The attachment window should be using the android:popupBackground attribute in the active theme. Check to make sure that is not set to transparent in your theme or your parent theme. If you have a specific background you'd like to use for that PopupWindow then you can set the drawable or color for android:popupBackground in your Theme.

lexi-sr:
Thank you! That was very helpful information, and I am able to get it working without this pull request.

PR #71, opened 2016-05-05
Added ability to add custom title method for conversations.

lexi-sr:
I just added the ability to set a custom title for the AtlasConversationsRecyclerView with as few changes as possible, and it will not break anyone's existing code either.

### DIFF
--- a/layer-atlas/src/main/java/com/layer/atlas/util/Util.java
+++ b/layer-atlas/src/main/java/com/layer/atlas/util/Util.java
@@ -54,6 +54,15 @@ public class Util {
     private static final String METADATA_KEY_CONVERSATION_TITLE = "conversationName";
     private static final int TIME_HOURS_24 = 24 * 60 * 60 * 1000;
     private static final SimpleDateFormat DAY_OF_WEEK = new SimpleDateFormat("EEE, LLL dd,", Locale.US);
+    private static CustomTitleMethod customTitleMethod = null;
+
+    public static void setCustomTitleMethod(CustomTitleMethod listenerToSet) {
+        customTitleMethod = listenerToSet;
+    }
+
+    public interface CustomTitleMethod {
+        String getConversationTitle(LayerClient client, ParticipantProvider provider, Conversation conversation);
+    }
 
     /**
      * Returns the app version name.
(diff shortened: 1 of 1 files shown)


smpete:
Thanks for the pull request. To improve generality, what do you think of changing this to set some Formatter object on the `AtlasConversationsAdapter`? There could be a default formatter that uses the methods found in `Util`. Then you can pass a custom `Formatter` in to the `AtlasConversationsRecyclerView` which would then set it on the adapter. This would probably need to be done in the constructor, else we'd have to possibly refresh the RecyclerView. Making it more extensible will allow time, last message string and other future strings to be customized easily.

lexi-sr:
Thanks for your response. I don't think the formatter will work for our particular use case. We have the conversation title stored in metadata (named by the user, any arbitrary name) but it is stored 2 levels deep.  For example, we have to use this code to get it:  `((Metadata) metadata.get("groupInfo")).get("name").`  So I could not just change `private static final String METADATA_KEY_CONVERSATION_TITLE = "conversationName";` in Util.java. The most direct solution for my case would be the ability to customize  `public static String getConversationMetadataTitle(Conversation conversation)`, so I can put the code that goes 2 levels deep?  Thanks!

smpete:
It sounds like we can extract both the `getConversationTitle` and `getConversationMetadataTitle` into the a new `ConversationFormatter` class. It could look something like this:
```java
public class ConversationFormatter {

    public String getConversationTitle(LayerClient client, ParticipantProvider provider, Conversation conversation) {
        String metadataTitle = getConversationMetadataTitle(conversation);
        if (metadataTitle != null) return metadataTitle.trim();

        StringBuilder sb = new StringBuilder();
        String userId = client.getAuthenticatedUserId();
        for (String participantId : conversation.getParticipants()) {
            if (participantId.equals(userId)) continue;
            Participant participant = provider.getParticipant(participantId);
            if (participant == null) continue;
            String initials = conversation.getParticipants().size() > 2 ? Util.getInitials(participant) : participant.getName();
            if (sb.length() > 0) sb.append(", ");
            sb.append(initials);
        }
        return sb.toString().trim();
    }

    public String getConversationMetadataTitle(Conversation conversation) {
        String metadataTitle = (String) conversation.getMetadata().get(METADATA_KEY_CONVERSATION_TITLE);
        if (metadataTitle != null && !metadataTitle.trim().isEmpty()) return metadataTitle.trim();
        return null;
    }
}
```

Then you can subclass to customize the functionality of the metadata title method and/or the conversation title method.

lexi-sr:
Ok, that would be great. Thanks!

Dan2anderson:
We've implemented those changes as requested and submitted a new pull request here https://github.com/layerhq/Atlas-Android/pull/167

Please close this pull request.  Thanks

PR #74, opened 2016-05-06
added list item selected indicator

Dan2anderson:
Added functionality to set default and selected background colors. These colors are used by the adapter to highlight an item when it is selected.  No highlighting occurs unless these colors are set. 

We need this feature for when we show a master detail flow view of the Conversations.  This way the user can see which item is selected in the list and not feel confused about what conversation's messages are showing in the details. 

Danny Anderson
Sales Rabbit.


smpete:
Thanks for the pull request. We should utilize the `activated` state on the item view instead of passing specific background resources. This would allow for modification at the theme level, as well as passing in drawable resources.

We should also generalize this by starting the implementation of choice modes in the `RecyclerView`s that Atlas uses, similar to how `ListView` works. This would allow for multiple selections and CAB support. For now we could restrict it to a single choice to make things simple.

Dan2anderson:
so we re-wrote our implementation of this feature to use the activated state as requested.  Please see https://github.com/layerhq/Atlas-Android/pull/166

We can close this pull request.

PR #76, opened 2016-05-10
Allow 3rd party to customise the background of the attachment menu

### LINKED ISSUE #75: The attachment menu's issues
* The current background is transparent
* Touching outside of the menu doesn't dismiss it

rockerhieu:
This PR also fixes #75


layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
```diff
@@ -75,6 +78,7 @@
     private int mTextStyle;
     private int mUnderlineColor;
     private int mCursorColor;
+    private int mAttachmentSendersBackgroundColor;
```
smpete: Let's remove the Color from the name so it denotes drawables will work too.

layer-atlas/src/main/res/values/atlas-attrs.xml
```diff
@@ -50,6 +50,7 @@
         <attr name="android:enabled" />
         <attr name="inputUnderlineColor" />
         <attr name="inputCursorColor" />
+        <attr name="attachmentSendersBackgroundColor" format="color|reference" />
```
smpete: Let's remove the Color from the name so it denotes drawables will work too. This would make it "attachmentSendersBackground".

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
```diff
@@ -337,6 +342,8 @@ private void initAttachmentMenu(Context context, AttributeSet attrs, int defStyl
         mAttachmentMenu.setContentView(LayoutInflater.from(context).inflate(R.layout.atlas_message_composer_attachment_menu, null));
         mAttachmentMenu.setWindowLayoutMode(ViewGroup.LayoutParams.WRAP_CONTENT, ViewGroup.LayoutParams.WRAP_CONTENT);
         mAttachmentMenu.setOutsideTouchable(true);
+        mAttachmentMenu.setBackgroundDrawable(new ColorDrawable(mAttachmentSendersBackgroundColor));
+        mAttachmentMenu.setFocusable(true);
```
smpete: Why does the popup window need to be focusable?

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
rockerhieu: But it has to be color, hasn't it? Otherwise this line will be error:
```java
this.mAttachmentSendersBackgroundColor = ta.getColor(R.styleable.AtlasMessageComposer_attachmentSendersBackgroundColor, context.getResources().getColor(R.color.atlas_color_primary_gray));
```

layer-atlas/src/main/res/values/atlas-styles.xml
```diff
@@ -47,6 +47,7 @@
         <item name="android:enabled">true</item>
         <item name="inputUnderlineColor">@color/atlas_color_primary_blue</item>
         <item name="inputCursorColor">@color/atlas_color_primary_blue</item>
+        <item name="attachmentSendersBackgroundColor">@color/atlas_color_primary_gray</item>
```
smpete: This value should be `@drawable/atlas_popup_background`

rockerhieu: Now I know what you meant. Will change to accept both drawable and color.

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
rockerhieu: So that clicking outside of the popup will dismiss it, and the outside touch will be consumed by the popup itself (not by any view at the touch). If the popup is not focusable, clicking on the `+` icon will first dismiss the popup and quickly show another one.

layer-atlas/src/main/res/layout/atlas_message_composer_attachment_menu_item.xml
```diff
@@ -17,6 +17,9 @@
               android:layout_width="match_parent"
               android:layout_height="@dimen/atlas_dropdown_item_height"
               android:orientation="horizontal"
+              android:background="?attr/selectableItemBackground"
+              android:paddingLeft="@dimen/atlas_padding_narrow"
```
smpete: Let's keep these as `atlas_padding_normal` instead of narrow.

smpete:
Can you also remove the `android:popupBackground` from the `atlas-themes.xml` file? Since we are no longer using the theme's popupBackground for the attachment menu this is no longer necessary. It will also fix the case of the background of the select text handles being incorrect when copying text.

rockerhieu:
@smpete done!

smpete:
Merged. Squashed the commit so it is only one commit into master.

PR #79, opened 2016-05-18
Set bitmaps to Avatars

lexi-sr:
I need the ability to set bitmaps to avatars. I was able to get this far, but then I wasn't sure how I would get the correct scaling on the bitmap. Here is a screenshot that demos the scaling problem with clustering & nonclustering: (see screenshot at 
[this link](https://drive.google.com/file/d/0B-l17HPY9SgHZTBJRjZrRVpZeUk/view?usp=sharing)).  I think fixing the scaling might fix the clustering problem.

P.S. This is the full-size avatar, for comparison: [link](https://drive.google.com/file/d/0B-l17HPY9SgHdW5lTjRWeWVzdjQ/view?usp=sharing)

How can I get the scaling correct? It is a little complicated to find the answer in the code. Thanks!


smpete:
What is the source of the bitmap? If it's a file (downloaded or in res/raw) then you can just create a `Uri` from that. If it is a drawable resource then you can add functionality to load that resource ID utilizing the existing implementation. The participant interface would need to have something similar to `int getAvatarResourceId()`. Then you can add setters/getters in the `AltasAvatar.ImageTarget` and load the resource using Picasso (see `AtlasAvatar.setClusterSizes()`).

We should eventually support error and placeholder drawables as well.

lexi-sr:
Thank you for your response. It's a byte array which was loaded from the web in a JSON object, and then saved in a database as a byte array, which then needs to be decoded into a bitmap. But we would like to continue using the initials if some users do not have a byte array (they have no profile picture).

smpete:
Rather than saving it to the database you could create the bitmap file when parsing the JSON and save the file path in the database instead. You may see increased performance depending on image sizes (see http://www.sqlite.org/intern-v-extern-blob.html). Then you can just use the file's `Uri` when setting the participant's avatar.

lexi-sr:
Thank you for your quick response. Thanks for the idea. I checked, and our maximum byte array size of the image stored in our database is 1400 KB, so according to that link, it is faster to store it in our database directly.

smpete:
Whoa, 1.4MB? It should be faster to store them in a file I presume. The link stated the following rule of thumb:

> For BLOBs larger than 100KB, reads from a separate file are faster.

lexi-sr:
Oops, I made a mistake interpreting the chart. I am currently waiting to hear back from my colleague to see what we would like to do moving forward. I'll get back to you. Thanks!

rakshitsoni02:
May i know how to set avatar and name for a user id to heroku as i am getting json array of user from  String url = "https://layer-identity-provider.herokuapp.com/apps/" + mLayerAppIdLastPathSegment + "/atlas_identities";

PR #81, opened 2016-05-25
Task/update to 0.21.0

smpete:


sriamar:
We have not been doing a good job of updating changelog.gradle and version. Let's do that going forward. So, could you please bump version, and update changelog here as well.

smpete:
Sure thing!

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Atlas Android
 
+## 0.2.9
```
sriamar: This should be `0.2.10`. I already shipped a `0.2.9` and failed to update these. So, lets do this.

1. Change this to `0.2.10`, and add `0.2.9` with statement `Updated to Layer Android SDK Version 0.20.4` 
2. Update version to `0.2.10`
3. Tag it to `0.2.10`

PR #82, opened 2016-05-27
APPS-2444 - Calling refresh when init'ing the conversations recycler …

smpete:
…view

Fixes initial query execution when conversation recycler view is in a ViewPager and gets a visibility changed call before init() is called.
Since the adapter is notified through a Hander message on the main thread, there won't be problems configuring the RecyclerView after the query runs.


smpete:
Still need to update the release version and change log but wanted to get

PR #94, opened 2016-07-13
Update play services

alistairsykes:
Update the play services version


smpete:
This has since been updated. Version 0.3.2 is using play service 9.6.1

PR #97, opened 2016-08-09
Remove requires permission tag

alistairsykes:
This tag is not correct as that method does not require the location permission, since it is being handled inside whether you have the permission or not. This tag gives developers an error lint which is not correct.

---

This change is [Reviewable](https://reviewable.io/reviews/layerhq/atlas-android/97)


alistairbrightec:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/layerhq/atlas-android/97)*

thecombatwombat:
@alistairsykes I just double checked. Removing the annotation leads to an error + running lint on that class shows no issue with that particular annotation. 

There are still some lint errors there, and going forward we'd like to iron them out, but this does not seem to be an issue. 

https://developer.android.com/studio/write/annotations.html#permissions

PR #113, opened 2016-10-15
Adding stackFromEnd attribute to AtlasMessagesRecyclerView.

josealfonsomora:
AtlasMessageRecyclerView stacks messages from the botton when true
and stacks messages fro top when false

Usage:

``````
<com.layer.atlas.AtlasMessagesRecyclerView
    android:id=@+id/messages_list
    android:layout_width=match_parent
    android:layout_height=match_parent
    app:stackFromEnd=false/>```
``````


smpete:
Thanks for the pull request. This looks good but there is one issue. Auto scroll to the bottom of the message list is broken. When displaying messages where there are more messages than the height of the screen the RecyclerView stays at the top and doesn't scroll to the newest message. Once this is fixed I would be happy to merge this.

josealfonsomora:
@smpete I wen through the autoscroll and found this piece of code
`private void autoScroll() {
        int end = mAdapter.getItemCount() - 1;
        if (end <= 0) return;
        int visible = findLastVisibleItemPosition();
        // -3 because -1 seems too finicky
        if (visible >= (end - 3)) {
            scrollToPosition(end);
        }
    }`

What is doing is only autoscroll if the user is seeing the last 3 messages, if there is more messages than the height of the screen and the user is not seeing any of the the latests 3 will not scroll. That is good to prevent the scroll if the user is reading old messages.

I can't see the problem with that and my change, probably I'm missing something?

Thanks!

smpete:
The problem is when you initially open the messages view it is not scrolled to the bottom. This has two issues:
 - It shows the oldest messages instead of the newest messages
 - When the autoScroll() method is called due to a new message, it will be a no-op since it is not at the bottom of the list

You can see this effect if you have a conversation that has enough messages that it requires a scroll to see them all.

josealfonsomora:
Thanks @smpete I saw the problem. Yes, the scroll remains in the top and I'm afraid we will need a big refactor to solve this situation.

I tried doing a manual scroll after layout the childs, but there is a jump when the list is opened because we can't scroll to bottom before the view is draw.

` getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
            @Override
            public void onGlobalLayout() {

                if(!mStackFromEnd && mAdapter.getItemCount() > 0) {
                    scrollToPosition(mAdapter.getItemCount()-1);
                    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
                        getViewTreeObserver().removeGlobalOnLayoutListener(this);
                    } else {
                        getViewTreeObserver().removeOnGlobalLayoutListener(this);
                    }
                }
            }
        });`

The concept in RecyclerView of stackFromEnd = false will not allow us to have the scroll at the bottom and unfortunately we can't know how many items do we have in the list in drawing time :-(

atlas_recycler_view

If we could know on initialisation, the number of childs, we could change the AtlasRecyclerView to be stacked from end or not, but that implies to couple the code. Or if we could hide the view until scrolled.. but I don't know how.. Any advice?

smpete:
We won't be able to know the size at initialization because the query is run asynchronously. Hiding the view until the data is loaded for the first time may be a viable solution but there has to be a cleaner way of doing it. I'll try to find some time in the next few days to work on a solution.

smpete:
I'm going to close this as we are making major changes to the internals of this recycler view over then next two months. This feature will be included in that work so expect availability in the coming months. If this is needed before that then please re-open.

Thanks!

PR #121, opened 2016-11-15
Added listener for when selecting a participant fails +fallback toast

thecombatwombat:


layer-atlas/src/main/java/com/layer/atlas/AtlasAddressBar.java
```diff
@@ -743,4 +758,8 @@ public ViewHolder(ViewGroup parent) {
     public interface OnParticipantSelectionChangeListener {
         void onParticipantSelectionChanged(AtlasAddressBar conversationLauncher, List<Identity> participants);
     }
+
+    public interface OnParticipantSelectionFailedListener {
+        void onParticipantSelectionFailed (String reason);
```
thecombatwombat: Okay, yeah that makes sense... Lemme look up a usage...

thecombatwombat:
@smpete @sriamar ready for you guys to take another look...

layer-atlas/src/main/java/com/layer/atlas/AtlasAddressBar.java
```diff
@@ -47,11 +48,14 @@
 import java.util.Set;
 
 public class AtlasAddressBar extends LinearLayout {
+    public static final int MAX_PARTICIPANTS = 25;
```
sriamar: Ideally, this should come from SDK. File a ticket in JIRA to track that.

thecombatwombat: 👍

PR #122, opened 2016-11-24
Remove dependence on Util.java for message preview text

econnelly:
This isn't ready for production, but I think this type of system is a lot more developer-friendly. I was recently integrating Atlas into an app and came across an issue where the preview text was always being generated by `GenericCellFactory`. There was no way for me to modify the preview text without importing the entire Atlas source and modifying `Util.java`. Even that wasn't straight forward because my custom cell factory was out of scope of the library.

I modified Atlas to allow for cell factory registration on init through a configuration class.

What this pull request does:
 - Creates an `Atlas` class to hold the `Configuration` class for registering cell factories
 - Uses annotations to determine the `isType()` and `getPreviewText()` methods
   - Custom cell factories need to create static methods annotated with `@IsType` and `@PreviewText` annotations
   - Reflection is used to find the annotated methods. It's an up-front cost at registration time. From then on it's just the cost to call `invoke()` when calling the methods.
 - Automatically creates instances of cell factories
   - This has a limitation of requiring that cell factories have a 0-argument constructor


econnelly:
Usage:

**In `onCreate` of the Application class**
```
Atlas.initialize(
new Configuration.Builder(this)
.addCellFactory(MyCustomCellFactory.class)
.build());
```

**MyCustomCellFactory.java**
```
public class MyCustomCellFactory extends AtlasCellFactory<MyCustomCellFactory.CellHolder, MyCustomCellFactory.CellInfo> {

    @IsType
    public static boolean isType(Message message) {
      ...
    }

    @PreviewText
    public static String getMessagePreview(Context context, Message message) {
     // context can be null
     ...
    }
...

}
```

smpete:
Thanks for the PR. We need to make this change but we'd prefer not to use reflection unless the performance overhead is non-existent in this case. I'm not certain off-hand how best to implement this but we may need to change `AtlasCellFactory` to lazily instantiate its `LruCache` if we need to store skeleton instances of the cell factories.

econnelly:
Custom cell factories will need to be registered somewhere. Either a full instance will need to be stored throughout the application's lifecycle, or just the class for use when needed. Lazy initializing the LruCache will provide a smaller footprint in memory, but there's a pretty large chance that the cell factory will never be used.

I've implemented both kinds of registries separately and haven't seen any noticeable decrease in performance from using reflection. This is also the system that ActiveAndroid uses. I sill have my other branch that registers cell factories by storing instances. I can do a different PR that does what you're suggesting.

thecombatwombat:
@econnelly please let us know if #124 fixes these problems, so I can square away this PR

PR #128, opened 2016-12-20
Using file content provider for CameraSender + added permissions conv…

thecombatwombat:
…enience methods


layer-atlas/src/main/java/com/layer/atlas/messagetypes/threepartimage/CameraSender.java
```diff
@@ -47,24 +56,52 @@ public CameraSender(String title, Integer iconResId, Activity activity) {
...
         return true;
     }
 
     @Override
+    public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+        if (requestCode != PERMISSION_REQUEST_CODE) return;
+
+        if (grantResults.length != 2) {
+            if (Log.isLoggable(Log.VERBOSE)) Log.v("One or more required  permissions denied");
+            return;
```
smpete: We should signal this to the developer somehow so they can show the necessary UI for the error, or file an improvement to do this as the GallerySender currently does not

layer-atlas/src/main/AndroidManifest.xml
```diff
@@ -19,5 +19,18 @@
 
     <application>
         <activity android:name=".util.imagepopup.AtlasImagePopupActivity"/>
+
+        <provider
+            android:authorities="${applicationId}.provider"
```
smpete: Will this cause conflicts if an app already has a provider with the defined authority?

thecombatwombat: Hmmm... possibly... I'll just define this under our namespace then...

PR #134, opened 2017-01-03
Fixed File URI exception that happens when try to take a picture from in the app. (on a nougat device).

Dan2anderson:
Fixed the CameraSender.startCameraIntent() so that it works on Android v7 (Uses the FileProvider  instead of File when v7 or higher).  Please note this fix uses the FileProvider of which you can only have one in your app.  So if any clients (like SalesRabbit) already uses the FileProvider they are goint to need to include this path in thier provider paths xml file '<external-path name="external_files" path="."/>'.   In the <provider> tag for your fileProvider you need to add tools:replace="android:authorities". In the <meta-data> tag for your the same provider you need to addtools:replace="android:resource"


thecombatwombat:
@Dan2anderson #140 already fixes this. it removes the `FileProvider` inside Atlas, and requires you to supply the FileProvider authority to the `CameraSender` instance. 

The [Changelog](https://github.com/layerhq/Atlas-Android/blob/master/CHANGELOG.md) details this and we've provided a [guide](https://github.com/layerhq/Atlas-Android/blob/master/docs/CameraSender.md) in the docs as well

PR #140, opened 2017-01-19
Feature/fix file provider merge

thecombatwombat:


docs/CameraSender.md
```diff
@@ -0,0 +1,26 @@
...
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/provider_paths"/>
+
+        </provider>
+```
+Create a `provider_paths.xml` file in the `xml` folder under `res` folder. Folder may be needed to create if it doesn't exist.
+
+And then, when supply the authority to the `CameraSender` when creating an instance.
```
smpete: Should be "And then, supply the authority to the `CameraSender` when creating an instance."

docs/CameraSender.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+new CameraSender(R.string.attachment_menu_camera,R.drawable.ic_photo_camera_white_24dp, mActivity ,getApplicationContext().getPackageName() + ".file_provider")
```
smpete: There is a comma after `mActivity`. Perhaps this should just be `{authority}` and mention that in this case with the example provider it is `mActivity.getApplicationContext().getPackageName() + ".file_provider"`. Mainly because they may use some other authority than `${applicationId}.file_provider`. Or mention to replace this authority with the one defined in their existing provider.

smpete: Speaking of, maybe mention to just pass their authority if they have an existing provider

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Atlas Android
 
+## 0.3.8
+
+### Bug Fixes
+  * Fixed manifest merge issues + crashes caused by workarounds by removing the FileProvider required for CameraSender. 
+  Developers are now requried to implement their own FileProvider and supply the CameraSender with the provider authority.
+
```
smpete: We should add a migration guide. Technically this should be `0.4.0` as well since this is a breaking API change.

docs/CameraSender.md
```diff
@@ -18,9 +18,12 @@ Atlas provides a simple way for you to take an image via the camera and send it
 ```
 Create a `provider_paths.xml` file in the `xml` folder under `res` folder. Folder may be needed to create if it doesn't exist.
 
-And then, when supply the authority to the `CameraSender` when creating an instance.
+And then, then supply the authority to the `CameraSender` when creating an instance.
```
smpete: Should this be just "Then supply the authority..."? "And then, then" sounds redundant.

docs/CameraSender.md
```diff
@@ -18,9 +18,12 @@ Atlas provides a simple way for you to take an image via the camera and send it
 ```
 Create a `provider_paths.xml` file in the `xml` folder under `res` folder. Folder may be needed to create if it doesn't exist.
```
smpete: "Folder may be needed..." should be changed to something like "The folder may be needed..."

docs/CameraSender.md
thecombatwombat: Okay...

PR #143, opened 2017-01-25
Handling remote files from gallery in >=Kitkat && <=Nougat

thecombatwombat:


layer-atlas/src/main/java/com/layer/atlas/messagetypes/threepartimage/ThreePartImageUtils.java
```diff
@@ -266,6 +292,27 @@ private static Message newThreePartImageMessageFromUri(Context context, LayerCli
         return client.newMessage(parts);
     }
 
+    private static Message newThreePartImageMessageFromFileDescriptor(Context context, LayerClient client, @NonNull FileDescriptor fileDescriptor) throws IOException {
+        if (client == null) throw new IllegalArgumentException("Null LayerClient");
+
+        InputStream inputStream = new FileInputStream(fileDescriptor);
+        String filePath = context.getCacheDir() + "/img_" + Calendar.getInstance().getTimeInMillis();
+        writeStreamToFile(filePath, inputStream);
```
smpete: Why do we need to copy the file? This should be documented as comments

layer-atlas/src/main/java/com/layer/atlas/messagetypes/threepartimage/ThreePartImageUtils.java
```diff
@@ -59,15 +65,37 @@ public static MessagePart getFullPart(Message message) {
     public static Message newThreePartImageMessage(Context context, LayerClient layerClient, Uri imageUri) throws IOException {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             return newThreePartImageMessageFromUri(context, layerClient, imageUri);
-        } else {
-            Cursor cursor = context.getContentResolver().query(imageUri, new String[]{MediaStore.MediaColumns.DATA}, null, null, null);
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
smpete: Can you add comments as to why we need special handling for the different version cases?

PR #149, opened 2017-02-14
Feature/optional oneonone avatar

thecombatwombat:
Currently the `AtlasMessagesAdapter` hides the avatar for the other participant if the conversation has just 2 participants, i.e. is a 1:1 conversation. 

The default is still to hide the avatar, but there is now a method available on `AtlasMessagesRecyclerView` that allows this behavior to be optional.

### DIFF
--- a/layer-atlas/src/main/java/com/layer/atlas/AtlasMessagesRecyclerView.java
+++ b/layer-atlas/src/main/java/com/layer/atlas/AtlasMessagesRecyclerView.java
@@ -196,6 +196,27 @@ public class AtlasMessagesRecyclerView extends RecyclerView {
     }
 
     /**
+     * Convenience pass-through to this list's AtlasMessagesAdapter.
+     *
+     * @see AtlasMessagesAdapter#getShouldShowAvatarInOneOnOneConversations()
+     */
+
+    public boolean getShouldShowAvatarInOneOnOneConversations() {
+        return mAdapter.getShouldShowAvatarInOneOnOneConversations();
+    }
+
+    /**
+     * Convenience pass-through to this list's AtlasMessagesAdapter.
+     *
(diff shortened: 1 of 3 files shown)


layer-atlas/src/main/java/com/layer/atlas/AtlasMessagesRecyclerView.java
```diff
@@ -196,6 +196,27 @@ public View getFooterView() {
     }
 
     /**
+     * Convenience pass-through to this list's AtlasMessagesAdapter.
```
smpete: We should either state what these methods do in the java doc or add java docs to the accessor/mutator in the `AtlasMessagesAdapter` that describes what this flag does.

layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
```diff
@@ -95,6 +95,8 @@
     private RecyclerView mRecyclerView;
     private boolean mReadReceiptsEnabled = true;
 
+    protected boolean mShouldShowAvatarInOneOnOneConversations;
```
smpete: This should also be a style setting that the user can set via the layout XML. Let's file a task for that and do it separately.

layer-atlas/src/main/java/com/layer/atlas/AtlasMessagesRecyclerView.java
thecombatwombat: Done

layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
thecombatwombat: created ticket

sriamar:
Please add a description to the PR.

PR #154, opened 2017-02-23
Added functionality so that conversation items in the list are highli…

Dan2anderson:
…ghted when selected.  Also added functionality so that we can programatically select a specific item in the list if needed.


Dan2anderson:
This is crap code I have made a new pull request for this same feature.

PR #166, opened 2017-03-29
Added functionality to select a chat item in the conversationsAdapter…

Dan2anderson:
In AtlasConversationAdapter added code that allows sets the activated to true when a user clicks on a recycler view item the adapter tracks the selected position so that the recycler view gets deactivated when it is used for a new item, and activated when the selected item gets put back in to a recycler view.  

Added a default selection method in AtlasConversationsRecyclerView and the AtlasConversationAdapter so that we can programatically select the first item (index 0)in the adapter  if needed.

### DIFF
--- a/layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
+++ b/layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
@@ -65,18 +65,17 @@ public class AtlasConversationsRecyclerView extends RecyclerView {
 
         mAdapter = new AtlasConversationsAdapter(getContext(), layerClient, picasso);
         mAdapter.setStyle(conversationStyle);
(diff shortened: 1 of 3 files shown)


thecombatwombat:
@Dan2anderson as far as I can tell, this code marks an item as "selected" upon clicking, which opens the messages view, and then when the user comes back to the conversations list, the item continues to be in the "selected" state. 

I'm not entirely sure what the use case is, esp when you consider that this change would affect all applications that use Atlas. I don't see a compelling case to accept the PR yet. 

I'm going to close this PR, but in case you feel this needs to be in Atlas, please do the following:

1. Update your branch by rebasing it off `develop` to ensure there are no conflicts. 
2. Make the PR against `develop`, since that allows us to fit it into our release workflow
3. Add an explanation of the use case you envision, perhaps with some screenshots?

Thanks,
Archit

PR #171, opened 2017-04-12
Expose xml attribute to allow user set if avatar 

andela-wakinsanmi:
Expose xml attribute to allow user set if avatar should be shown in One on One conversation.

### DIFF
--- a/layer-atlas/src/main/res/values/atlas-attrs.xml
+++ b/layer-atlas/src/main/res/values/atlas-attrs.xml
@@ -64,7 +64,7 @@
         <attr name="theirTextSize" />
         <attr name="theirTextTypeface" />
         <attr name="theirTextStyle" />
-        <attr name="shouldShowAvatar" format="boolean" />
+        <attr name="setShouldShowAvatarsInOneOnOneConversations" format="boolean" />
     </declare-styleable>
 
     <declare-styleable name="AtlasAddressBar">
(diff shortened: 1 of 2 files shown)


layer-atlas/src/main/res/values/atlas-attrs.xml
```diff
@@ -64,6 +64,7 @@
         <attr name="theirTextSize" />
         <attr name="theirTextTypeface" />
         <attr name="theirTextStyle" />
+        <attr name="shouldShowAvatar" format="boolean" />
```
thecombatwombat: This name should be `mShouldShowAvatarInOneOnOneConversations`

layer-atlas/src/main/res/values/atlas-attrs.xml
```diff
@@ -64,7 +64,7 @@
         <attr name="theirTextSize" />
         <attr name="theirTextTypeface" />
         <attr name="theirTextStyle" />
-        <attr name="shouldShowAvatar" format="boolean" />
+        <attr name="setShouldShowAvatarsInOneOnOneConversations" format="boolean" />
```
thecombatwombat: @andela-wakinsanmi the xml property should be called `shouldShowAvatarsInOneOnOneConversations`.
The class member variable should be called `shouldShowAvatarsInOneOnOneConversations`. 
The setter of that member variable should be called `setShouldShowAvatarsInOneOnOneConversations()`.
The getter of that member variable should be called `getShouldShowAvatarsInOneOnOneConversations`

andela-wakinsanmi: Thanks @thecombatwombat I have changed the xml property to shouldShowAvatarsInOneOnOneConversations, I bet you mean the member variable should be mShouldShowAvatarsInOneOnOneConversations which is already in the branch

Issue #172, opened 2017-04-12
NPE: SinglePartImageCellFactory starts AtlasImagePopupActivity with null fullId


explodes:
If the `fullId` extra is not set in `AtlasImagePopupActivity` then the Activity will fail to start because `ImageSource.uri(Uri)` will throw an NPE if the uri is null.

This trace was taken from logging:
```java
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.foo.app.android/com.layer.atlas.util.imagepopup.AtlasImagePopupActivity}: java.lang.NullPointerException: Uri must not be null
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2419)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2479)
       at android.app.ActivityThread.-wrap11(ActivityThread.java)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1347)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:148)
       at android.app.ActivityThread.main(ActivityThread.java:5420)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
Caused by java.lang.NullPointerException: Uri must not be null
       at com.davemorrissey.labs.subscaleview.ImageSource.uri(ImageSource.java:111)
       at com.layer.atlas.util.imagepopup.AtlasImagePopupActivity.onCreate(AtlasImagePopupActivity.java:83)
       at android.app.Activity.performCreate(Activity.java:6280)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1107)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2372)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2479)
       at android.app.ActivityThread.-wrap11(ActivityThread.java)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1347)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:148)
       at android.app.ActivityThread.main(ActivityThread.java:5420)
       at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #173, opened 2017-04-12
SinglePartImageCellFactory handles Message data more defensively to prevent NPE

explodes:


thecombatwombat:
@explodes Just saw #172 and we'd normally want to have steps to reproduce before we attempt a fix. 
Allow me to make a couple of suggestions... 

1. Get in touch with support using the steps [here](https://docs.layer.com/sdk/android/support)
We'd want to be sure there isn't a deeper issue at play. AFAIK a message part should always have an id. 

2. Make this PR against `develop`. That way, in case we decide this is the best solution, this fix will go into our normal workflow and should be out with the next release.

Thank you!
Cheers,
A

explodes:
Terribly sorry, my long winded description must not have gotten submitted. I'll resubmit this from develop.

PR #176, opened 2017-04-19
Bug/apps 3016 null avatar

sriamar:


sriamar:
Archit review it with me.

PR #183, opened 2017-05-19
Bug/apps 982 chat bubble is empty on receiving long message

andela-wakinsanmi:


layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -47,7 +59,7 @@ public CellHolder createCellHolder(ViewGroup cellView, boolean isMe, LayoutInfla
     @Override
     public TextInfo parseContent(LayerClient layerClient, Message message) {
         MessagePart part = message.getMessageParts().get(0);
-        String text = part.isContentReady() ? new String(part.getData()) : "";
+        String text = (part.isContentReady()) ? new String(part.getData()) : null;
```
smpete: Redundant parenthesis

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -2,27 +2,39 @@
...
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
 public class TextCellFactory extends AtlasCellFactory<TextCellFactory.CellHolder, TextCellFactory.TextInfo> implements View.OnLongClickListener {
     public final static String MIME_TYPE = "text/plain";
+    private ArrayList<LayerProgressListener> mListeners;
+    private LayerProgressListener mLayerProgressListener;
+    //This is used to bind TextView  to the exact message to ensure the right TextView is updated
+    private HashMap<TextView, Uri> mTextViewUriHashMap;
```
smpete: Perhaps this should use weak references to the TextViews so we don't have to worry about removing all the listeners in the messages adapter. Avoiding the instance of checks in the adapter would be nice.

Also we should code to the interface (i.e. be a `Map` instead of a `HashMap`)

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -60,11 +72,71 @@ public TextInfo parseContent(LayerClient layerClient, Message message) {
 
     @Override
     public void bindCellHolder(CellHolder cellHolder, final TextInfo parsed, Message message, CellHolderSpecs specs) {
+
+        //Checking if the TextView is being recycled, replace the value in the map with the new message id
+        if (!mTextViewUriHashMap.isEmpty() && mTextViewUriHashMap.containsKey(cellHolder.mTextView)) {
```
smpete: Don't need isEmpty check _and_ containsKey check. They are both constant time so no need to reduce readability.

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -60,11 +72,71 @@ public TextInfo parseContent(LayerClient layerClient, Message message) {
...
+            mTextViewUriHashMap.put(cellHolder.mTextView, message.getId());
+        }
+
+        //This string will be null if the message part content is not Ready
+        if (parsed.getString() == null) {
+            downloadMessage(message, cellHolder.mTextView);
```
smpete: The in-progress state seems funky here as it will just be an empty bubble. What does the UX look like when an image is downloading? It's a bubble that matches the dimens of the image right?

layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
```diff
@@ -79,6 +82,9 @@
     protected final Map<AtlasCellFactory, Integer> mMyViewTypesByCell = new HashMap<AtlasCellFactory, Integer>();
     protected final Map<AtlasCellFactory, Integer> mTheirViewTypesByCell = new HashMap<AtlasCellFactory, Integer>();
 
+    private ArrayList<LayerProgressListener> mTextCellFactoryDownloadListeners = new ArrayList<>();
```
smpete: Why are we storing a reference to these listeners?

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -60,11 +72,71 @@ public TextInfo parseContent(LayerClient layerClient, Message message) {
...
+                unregisterListener(this);
+            }
+        };
+
+        part.download(mLayerProgressListener);
+        if (mListeners != null) mListeners.add(mLayerProgressListener);
+    }
+
+    private void unregisterListener(LayerProgressListener layerProgressListener) {
+        mListeners.remove(layerProgressListener);
```
smpete: I don't understand why we need a reference to these listeners... This has no effect on the listener getting called by the SDK. It's just removing the reference from the member variable.

layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
andela-wakinsanmi: I thought about keeping the reference to the listeners in case we have more than one message that's being downloaded at the same time. This way, I can set the pointer to null in onDestroy. Is there a better way I can clear the reference to avoid memory leak?

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
andela-wakinsanmi: Yes it has no effect on the listener in the sdk, I assume the sdk will unregister the listener when the download is complete, I remove it from the list so that I can clear the map when the list is empty. I am thinking about scenario where we have more than one message being downloaded. Do you recommend I remove the list?

layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
smpete: The adapter shouldn't have to know about the implementation details in the TextCellFactory. That should all be handled at the factory level

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
smpete: But what is the point of the `mListeners` list? You're just adding/removing references, never fetching from that list. Can't you just not store the progress listeners as member variables to get the same effect?

> I assume the sdk will unregister the listener when the download is complete
Looking at the SDK code that is an incorrect assumption and it also appears that the three part image is doing it wrong as well. The progress listener is responsible for unregistering itself. It looks like you'd have to go through `LayerClient` to do that though...

layer-atlas/src/main/java/com/layer/atlas/adapters/AtlasMessagesAdapter.java
andela-wakinsanmi: Thanks @smpete. Thought about that but the blocker was how to notify the TextCellFactory the view onDestroy is being called so it can clear the reference.

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
smpete: This is still outstanding

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -60,11 +66,51 @@ public TextInfo parseContent(LayerClient layerClient, Message message) {
...
+                Uri uriValueInMap = mTextViewUriHashMap.get(textView);
+                if (uriValueInMap != null && uriValueInMap.equals(messageId) ) {
+                    textView.setText(new String(part.getData()));
+                    mTextViewUriHashMap.remove(textView);
+                }
+            }
+
+            @Override
+            public void onProgressError(MessagePart messagePart, Operation operation, Throwable throwable) {
+                mTextViewUriHashMap.remove(textView);
```
smpete: We should log this error

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -60,11 +68,63 @@ public TextInfo parseContent(LayerClient layerClient, Message message) {
...
+                if (cellHolder.mProgressBar.isShown()) {
+                    cellHolder.mProgressBar.hide();
+                }
+            }
+
+            @Override
+            public void onProgressError(MessagePart messagePart, Operation operation, Throwable throwable) {
+                mTextViewUriHashMap.remove(textView);
+                cellHolder.mProgressBar.hide();
+                Log.e(throwable.getMessage(), throwable);
```
smpete: This need s a log level check. Also use a more descriptive message than just the message from the throwable. That will be logged regardless since you're passing the throwable to the log

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -60,11 +68,63 @@ public TextInfo parseContent(LayerClient layerClient, Message message) {
...
+            public void onProgressComplete(MessagePart messagePart, Operation operation) {
+                //Check the downloaded message to ensure the TextView has not been recycled
+                Uri messageId = messagePart.getMessage().getId();
+                Uri uriValueInMap = mTextViewUriHashMap.get(textView);
+                if (uriValueInMap != null && uriValueInMap.equals(messageId) ) {
+                    textView.setText(new String(part.getData()));
+                    mTextViewUriHashMap.remove(textView);
+                }
+                if (cellHolder.mProgressBar.isShown()) {
+                    cellHolder.mProgressBar.hide();
```
smpete: This will hide even if this text view has been recycled and is no longer associated with the message part that was completed, right?

andela-wakinsanmi: Yes

andela-wakinsanmi: Thanks for pointing that out. Although, I have tested it, the progress bar doesn't show. I will take care of the condition by hiding the progress bar ```if (mTextViewUriHashMap.containsKey(cellHolder.mTextView)) {
            mTextViewUriHashMap.put(cellHolder.mTextView, message.getId());
            hideProgressbar(cellHolder);
        }```
This is when the view is being recycled.

PR #218, opened 2017-06-28
Feature/and 986/update avatar to layer ui

andela-wakinsanmi:


layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -0,0 +1,343 @@
...
+        }
+    }
+
+
+    private void drawPresence(Canvas canvas, Identity identity) {
+        Presence.PresenceStatus currentStatus = identity.getPresenceStatus();
+        if (currentStatus == null) {
+            return;
+        }
+        mViewModel.checkPresence(currentStatus, canvas);
```
smpete: I'm curious why the canvas needs to be passed back to the view model here.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    @Override
+    public void drawAvailable(Canvas canvas) {
```
smpete: It seems like these methods should not take a canvas in... Or is that required for image caching?

layer-atlas/src/main/java/com/layer/ui/util/imagecache/PicassoImageCacheWrapper.java
```diff
@@ -0,0 +1,107 @@
...
+                .build();
+
+        mTargets = new HashSet<>();
+    }
+
+    @Override
+    public void load(String targetUrl, String tag, int size, int size1, ImageView imageView, Object... args) {
+        boolean isMultiTransform = false;
+        if (args != null && args.length > 0) {
+            isMultiTransform = (boolean) args[0];
```
smpete: Type validation?

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageCacheWrapper.java
```diff
@@ -0,0 +1,23 @@
...
+import android.graphics.Bitmap;
+import android.widget.ImageView;
+
+public interface ImageCacheWrapper {
+    //
+    void load(String targetUrl, String tag, int width, int height, ImageView imageView, Object... args);
+
+    void fetchBitmap(String url, Object tag, int width, int height, final Callback callback, Object... args);
+
+    interface Callback {
```
thecombatwombat: Minor formatting comment. Put the interface after all the methods

layer-atlas/src/test/java/com/layer/ui/avatar/AvatarViewViewModelTest.java
```diff
@@ -0,0 +1,44 @@
...
+    Identity mMockIdentity, mMockIdentity2;
+
+    @Mock
+    Avatar.View mMockView;
+
+    AvatarViewModel mAvatarViewModel;
+
+    @Mock
+    ImageCacheWrapper mMockImageCacheWrapper;
+    @Mock Canvas mMockCanvas;
```
thecombatwombat: @andela-wakinsanmi pick a style for annotations, either on the same line or the next, and we should stick to that

layer-atlas/src/androidTest/java/com/layer/atlas/view/AvatarActivityTestView.java
```diff
@@ -0,0 +1,29 @@
+package com.layer.atlas.view;
```
sriamar: Why is this Atlas?

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -1,47 +1,33 @@
...
- *
+ * <p>
  * AtlasAvatar uses Picasso to render the avatar image. So, you need to init
  */
-public class Avatar extends View {
-    public static final String TAG = Avatar.class.getSimpleName();
-
-    private final static CircleTransform SINGLE_TRANSFORM = new CircleTransform(TAG + ".single");
-    private final static CircleTransform MULTI_TRANSFORM = new CircleTransform(TAG + ".multi");
+public class AvatarView extends View implements Avatar.View{
```
sriamar: space before `{`

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,225 @@
...
+
+
+    }
+
+    private String getInitialsForAvatarView(Identity added) {
+        return mAvatarInitials != null ? mAvatarInitials.getInitials(added) : Util.getInitials(added);
+    }
+
+    @Override
+    public void setParticipants(Identity[] participants) {
```
sriamar: Why do we need 2 setParticipants (one with array, and one with Set)?

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private Set<Identity> mParticipants = new LinkedHashSet<>();
+    private final Map<Identity, String> mInitials = new HashMap<>();
+    private final Map<Identity, BitmapWrapper> mImageTargets = new HashMap<>();
+    // Initials and Picasso image targets by user ID
+    private final List<BitmapWrapper> mPendingLoads = new ArrayList<>();
+    private AvatarInitials mAvatarInitials;
+
+
+    private Avatar.View mView;
```
sriamar: ViewModel shouldn't have a hard dependency on View. What if there are multiple views. What happens if View got destroyed?

sriamar: - This object is not used thread safe. 
- Also, ViewModel has a hard dependency on View (Circular reference). 

I think this should be a weakreference (https://developer.android.com/reference/java/lang/ref/WeakReference.html). And, should be handled in a thread safe way across the board.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
andela-wakinsanmi: Thanks @sriamar, ConversationAdapter and AvailableConversationAdapter in AddressBar call conversation.getParticipants() which returns a set of Identity. The one with Array is used with just an Identity. That's the reason for the method overload, any idea on how can I make this better?

thecombatwombat: This is just legacy code. We can fix it later

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
andela-wakinsanmi: I will change it to WeakReference, is there anything else that needs to be done in respect to this  comment`This object is not used thread safe`

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
sriamar: Okay, sounds good!

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,225 @@
...
+    @Override
+    public void loadImage(String url, String tag, int width, int height, final BitmapWrapper bitmapWrapper, Object... args) {
+
+        mImageCacheWrapper.fetchBitmap(url, tag, width, height,
+                new ImageCacheWrapper.Callback() {
+                    @Override
+                    public void onSuccess(Bitmap bitmap) {
+                        if (mView != null) {
+                            bitmapWrapper.setBitmap(bitmap);
+                            mView.revalidateView();
```
thecombatwombat: @andela-wakinsanmi lets make sure `mView` is a weak reference and its value is fetched into a local variable and null checked before use. Please make that change for everywhere mView is used.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,225 @@
...
+                        }
+                    }
+
+                    @Override
+                    public void onFailure() {
+                        bitmapWrapper.setBitmap(null);
+                        mView.revalidateView();
```
thecombatwombat: @andela-wakinsanmi  lets make sure mView is a weak reference and its value is fetched into a local variable and null checked before use. Please make that change for everywhere mView is used.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,225 @@
...
+                    }
+
+                    @Override
+                    public void onPrepareLoad() {
+                        bitmapWrapper.setBitmap(null);
+                        mView.revalidateView();
```
thecombatwombat: @andela-wakinsanmi  lets make sure mView is a weak reference and its value is fetched into a local variable and null checked before use. Please make that change for everywhere mView is used.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -114,10 +115,10 @@ public void update() {
         mPendingLoads.clear();
         mPendingLoads.addAll(toLoad);
 
-        mView.setClusterSizes(mInitials,mPendingLoads);
-        mView.revalidateView();
-
-
+        if (mView != null && mView.get() != null) {
+            mView.get().setClusterSizes(mInitials,mPendingLoads);
```
thecombatwombat: @andela-wakinsanmi In all these cases, `mView.get` needs to be assigned to a local variable, and then the local variable should be used to make the required method calls. 

``` java
Avatar.View view = mView !=null ? mView.get() : null;
if (view!=null) {
    view.setClusterSizes(mInitials, mPendingLoads);
    view.revalidateView();
}
```

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
thecombatwombat: @andela-wakinsanmi For now, you can just wrap up calls to the view in a synchronized block

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -45,7 +45,7 @@ public void setAvatarInitials(AvatarInitials avatarInitials) {
     }
 
     @Override
-    public void update() {
+    public synchronized void update() {
```
thecombatwombat: @andela-wakinsanmi the synchronization should be on the `view` object, not on the methods. 

When synchronization is on the method, you're synchronizing on the object that the method is on. We want to synchronize on the `view`, not on the `AvatarViewModel` instance

thecombatwombat: @sriamar please correct me if I am wrong about ☝️

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -307,9 +219,15 @@ private boolean setClusterSizes() {
     }
 
     @Override
+    public void revalidateView() {
+        // Invalidate the current view, so it refreshes with new value.
+        postInvalidate();
```
andela-wakinsanmi: @sriamar This revalidateView method is called from the ViewModel.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,262 @@
...
+            toLoad.add(existingTarget);
+        }
+
+        for (BitmapWrapper bitmapWrapper : mPendingLoads) {
+            mImageCacheWrapper.cancelRequest(bitmapWrapper.getUrl());
+        }
+        mPendingLoads.clear();
+        mPendingLoads.addAll(toLoad);
+
+        if (mViewWeakReference != null) {
```
sriamar: You don't need synchronized etc here. All you need is this.

```
Avater.View view = mViewWeakReference.get();
if (view != null) {
    view.setClusterSizes(mInitials, mPendingLoads);
    view.revalidateView();
}
```

sriamar: Also, did you actually mean to call `view.setClusterSizes()` or `this.setClusterSizes()` here?

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+ * @see Avatar.View is implemented by
+ * @see AvatarView
+ **/
+
+public interface Avatar {
+
+    interface ViewModel {
+
+        //Loop through the participants and calls setClusterSizes and revalidateView() on the View
+        void update();
```
sriamar: I don't think we need update(). I see this is used in View's `setParticipants()`. That should do the work internally to do the work to update view. View need not have to explicitly call `udpate()`

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+        void setParticipants(Identity[] participants);
+
+        void setParticipants(Set<Identity> participants);
+
+        Set<Identity> getParticipants();
+
+        //Get the number of Avatar to be drawn in the View onDraw
+        int getInitialSize();
+
+        Set<Map.Entry<Identity, String>> getEntrySet();
```
sriamar: Remove `getInitialSize()` and `getEntrySet()`. Then, call it `HashMap<> getIdentityInitials()`. You can then do a count for getting size. You can get `entrySet` from HashMap if you want that.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+
+        //BitmapWrapper is an interface that wraps the bitmap
+        BitmapWrapper getImageTarget(Identity key);
```
sriamar: `getImageTarget` should be `getBitmapWrapper`

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,237 @@
...
+
+            BitmapWrapper existingTarget = mImageTargets.get(existing);
+            mImageCacheWrapper.cancelRequest(existingTarget.getUrl());
+            toLoad.add(existingTarget);
+        }
+
+        for (BitmapWrapper bitmapWrapper : mPendingLoads) {
+            mImageCacheWrapper.cancelRequest(bitmapWrapper.getUrl());
+        }
+        mPendingLoads.clear();
```
sriamar: This will lead to problems. You should calculate the diff between `toLoad` & `mPendingLoads`.

- Only cancel those that are not needed any more.
- queue up new request for all those that are needed now

Otherwise, you are cancelling & requing download everytime `update()` is called. (which is a lot, as this is called whenever view is refreshed)

sriamar: Reading more, this is from the old code. So, I am okay with not changing it right away, but do realize that this could lead to performance problems.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+
+        //setClusterSizes is called in the View onLayout which in turns call the corresponding method on the View
+        void setClusterSizes();
```
sriamar: This method should be removed.  Here is the current flow.
View::onLayout() -> ViewModel::setClusterSizes() -> ViewModel::updateView() -> View::setClusterSizes(mInitials, mPendingLoads).

Instead, it should be View::onLayout() -> View::setClusterSizes(). 

But, in View::setClusterSizes(), you need to mInitials & mPendingLoads.
- You can get mInitials via ViewModel::getIdentityInitials()
- Add a new method to ViewModel Interface called ViewModel::getBitmapWrappers(), that returns mPendingLoads. 
Now, you have all you need to do the work in View::setClusterSizes()

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
sriamar: Okay, reading further more, I think this is what is causing your problems with images loading slowly. If you do the diff I suggest above, your perf problems might just go away.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+
+         //loadImage Work with Image Caching Library to provide Bitmap to the View
+        void loadImage(String url, String tag, int width, int height, BitmapWrapper bitmapWrapper, Object... args);
+
+        //Set the view on the ViewViewModel
```
sriamar: `viewViewModel` should be `viewModel`?

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+        void setView(Avatar.View avatar, Handler handler);
```
sriamar: - Can we make this a `weakReference` parameter for both View & Handler?
- Can we call the method setViewAndHandler(...)

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarInitials.java
```diff
@@ -0,0 +1,7 @@
+package com.layer.ui.avatar;
+
+import com.layer.sdk.messaging.Identity;
+
+public interface AvatarInitials {
+    String getInitials(Identity added);
```
sriamar: Please change `added` should be `identity`?

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
andela-wakinsanmi: Ok @sriamar I will do that

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,237 @@
...
+        mPendingLoads.addAll(toLoad);
+
+        updateView(null, null, true);
+    }
+
+    private String getInitialsForAvatarView(Identity added) {
+        return mAvatarInitials != null ? mAvatarInitials.getInitials(added) : Util.getInitials(added);
```
sriamar: Implement a `AvatarInitialsImpl.java`. Then, set that using `setAvatarInitials()`. 

In AvaterInitialsImpl.java's getInitials(), call `Util.getInitials()`.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+
+        //BitmapWrapper is an interface that wraps the bitmap
+        BitmapWrapper getImageTarget(Identity key);
+
+        //setClusterSizes is called in the View onLayout which in turns call the corresponding method on the View
+        void setClusterSizes();
+
+         //loadImage Work with Image Caching Library to provide Bitmap to the View
+        void loadImage(String url, String tag, int width, int height, BitmapWrapper bitmapWrapper, Object... args);
+
```
sriamar: if load() is part of interface, cancel should be too?!

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+        //Set the view on the ViewViewModel
+        void setView(Avatar.View avatar, Handler handler);
+
+        //Set custom AvatarInitial on the ViewModel to allow client plug in their custom Initials
+        void setAvatarInitials(AvatarInitials avatarInitials);
+
+        void setMaximumAvatar(int maximumAvatar);
```
sriamar: This made me think a lot. 
- Everything in ViewModel::update() should be done in View.
- That means, you have to create a new method `ViewModel::getImageCacheWrapper()`. Then hold on to the weak reference in View, and directly use that from `View` for loading / cancelling, etc.
- ViewModel also doesn't have to maintain mPendingLoads.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    interface View {
+
+        boolean setClusterSizes(Map<Identity, String> initials, List<BitmapWrapper> pendingLoads);
```
sriamar: Based on above comments, we don't need this method.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
andela-wakinsanmi: This setMaximumAvatar is exposed through XML so that the client can specify the maximum Avatars to be drawn. Previously it was a constant but a TODO comment to move it to XML was on it.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarInitials.java
```diff
@@ -0,0 +1,7 @@
+package com.layer.ui.avatar;
+
+import com.layer.sdk.messaging.Identity;
+
+public interface AvatarInitials {
```
thecombatwombat: Rename this to something that reflects a more general use case, like `IdentityNameFormatter`

layer-atlas/build.gradle
```diff
@@ -63,6 +63,9 @@ dependencies {
     compile 'com.android.support:appcompat-v7:24.2.1'
     compile 'com.android.support:design:24.2.1'
 
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.mockito:mockito-core:2.+'
```
thecombatwombat: @andela-wakinsanmi ☝️

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
thecombatwombat: @sriamar The only place `revalidate` is called from is `updateView`, and all that revalidate does is call `postInvalidate()` on the view. 

`postInvalidate` is needed when you're trying to invalidate the view from a non UI thread. Which begs the question, why would updateView be called on any other thread? 

`revalidate`'s functionality can be achieved by simply calling `((View)mView).invalidate()`

sriamar: Yes!

sriamar: Then, remove the `view` interface, and just pass in android view in `setView` ?

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,33 @@
...
+ **/
+
+public interface Avatar {
+
+    interface ViewModel {
+
+         //loadImage Work with Image Caching Library to provide Bitmap to the View
+        void loadImage(String url, String tag, int width, int height, BitmapWrapper bitmapWrapper, Object... args);
+
+        void cancelImage(String url);
```
sriamar: Remove `loadImage` and `cancelImage`? They just redirect calls in `ViewModel` to ImageCacheWrapper. We already have `getImageCacheWrapper` in Interface? Why not remove these 2, and directly use that in `View`?

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+    public String getInitialsForAvatarView(Identity added) {
+        return mIdentityNameFormatter.getInitials(added);
+    }
+
+    @Override
+    public IdentityNameFormatter getIdentityNameFormatter() {
+        return mIdentityNameFormatter;
+    }
+
+    @Override
```
sriamar: Remove this indirection here. View already has access to ImageCacheWrapper. Just let it call this directly.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+    public void loadImage(String url, String tag, int width, int height, final BitmapWrapper bitmapWrapper, Object... args) {
+
+        mImageCacheWrapper.fetchBitmap(width, height, bitmapWrapper, args);
+    }
```
sriamar: Remove this indirection here. View already has access to ImageCacheWrapper. Just let it call this directly.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public ImageCacheWrapper getImageCacheWrapper() {
+        return mImageCacheWrapper;
+    }
+
+    @Override
+    public void cancelImage(String url) {
+        mImageCacheWrapper.cancelRequest(url);
+    }
```
sriamar: Remove this cancelImage here. View already has access to ImageCacheWrapper. Just let it call this directly.

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageCacheWrapper.java
```diff
@@ -0,0 +1,14 @@
+package com.layer.ui.util.imagecache;
+
+import android.widget.ImageView;
+
+public interface ImageCacheWrapper {
+
+    void load(String targetUrl, String tag, int width, int height, ImageView imageView, Object... args);
```
sriamar: Let's remove unused `load` method.

layer-atlas/src/main/java/com/layer/ui/util/imagecache/PicassoImageCacheWrapper.java
```diff
@@ -0,0 +1,74 @@
...
+import static com.layer.ui.util.Log.TAG;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class PicassoImageCacheWrapper implements ImageCacheWrapper {
+    protected final static CircleTransform SINGLE_TRANSFORM = new CircleTransform(TAG + ".single");
+    protected final static CircleTransform MULTI_TRANSFORM = new CircleTransform(TAG + ".multi");
+    protected final Picasso mPicasso;
+    private Set<Target> mTargets;
```
sriamar: You don't need mTargets, as it is not used anywhere?

andela-wakinsanmi: Thanks @sriamar Picasso keeps a WeakReference to load(Target), if we don't hold on strong reference to it, the request is cancelled by Picasso. We can use Fetch, it will only force the download and not provide us with Bitmap on Success. https://github.com/square/picasso/issues/83

layer-atlas/src/androidTest/java/com/layer/ui/test/GenericCellFactoryTest.java
```diff
@@ -1,11 +1,18 @@
-package com.layer.atlas.test;
+package com.layer.ui.test;
 
 import android.content.Context;
 import android.support.test.InstrumentationRegistry;
 
+<<<<<<< HEAD:layer-atlas/src/androidTest/java/com/layer/atlas/test/GenericCellFactoryTest.java
```
sriamar: Merge problem?

sriamar: Same below in this file.

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -729,10 +732,10 @@ public CellViewHolder(View itemView, Picasso picasso, boolean shouldShowAvatarPr
             mCell = (ViewGroup) itemView.findViewById(R.id.cell);
             mReceipt = (TextView) itemView.findViewById(R.id.receipt);
 
-            mAvatar = ((Avatar) itemView.findViewById(R.id.avatar));
-            if (mAvatar != null)  {
-                mAvatar.init(picasso);
-                mAvatar.setShouldShowPresence(shouldShowAvatarPresence);
+            mAvatarView = ((AvatarView) itemView.findViewById(R.id.avatar));
+            if (mAvatarView != null)  {
+                mAvatarView.init(new AvatarViewModel(imageCachWrapper), new IdentityNameFormatterImpl());
```
sriamar: This means that AvatarViewModel's life is controlled by AvatarView. It shouldn't be. It should be tied to life of MessagesAdapter here (?). So, make AvatarViewModel a member of MessagesAdapter now, and pass in the reference to AvatarView. When we do the work on Message to seperate it, things will become more clear.

Otherwise, AvatarViewModel will get destroyed when View goes out of scope. The whole goal is to avoid that.

layer-atlas/src/main/java/com/layer/ui/adapters/ConversationsAdapter.java
```diff
@@ -149,7 +175,7 @@ public ConversationsAdapter setOnConversationClickListener(OnConversationItemCli
     @Override
     public ViewHolder onCreateViewHolder(final ViewGroup parent, int viewType) {
         UiConversationItemBinding binding = UiConversationItemBinding.inflate(mInflater, parent, false);
-        binding.avatar.init(mPicasso);
+        binding.avatar.init(new AvatarViewModel(mImageCacheWrapper), new IdentityNameFormatterImpl());
```
sriamar: This means that AvatarViewModel's life is controlled by AvatarView. It shouldn't be. It should be tied to life of this class (ConversationsAdapter). So, make AvatarViewModel a member of this class now, and pass in the reference to AvatarView. When we do the work on other UI items to separate them, things will become more clear.

Otherwise, AvatarViewModel will get destroyed when View goes out of scope. The whole goal is to avoid that.

layer-atlas/src/main/java/com/layer/ui/AddressBar.java
```diff
@@ -455,9 +458,9 @@ public ParticipantChip(Context context, Identity participant, Picasso picasso) {
 
             // Initialize participant data
             mName.setText(Util.getDisplayName(participant));
-            mAvatar.init(picasso)
-                    .setParticipants(participant);
-            mAvatar.setStyle(mAvatarStyle);
+            mAvatarView.init(new AvatarViewModel(mImageCacheWrapper), new IdentityNameFormatterImpl())
```
sriamar: This means that AvatarViewModel's life is controlled by AvatarView. It shouldn't be. It should be tied to life of this class (AddressBar). So, make AvatarViewModel a member of this class now, and pass in the reference to AvatarView. When we do the work on other UI items to separate them, things will become more clear.

layer-atlas/src/main/java/com/layer/ui/AddressBar.java
```diff
@@ -608,10 +609,10 @@ private void queryConversations(Set<Identity> selectedParticipants) {
         @Override
         public ViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
             ViewHolder viewHolder = new ViewHolder(parent);
-            viewHolder.mAvatar
-                    .init(mPicasso);
-            viewHolder.mAvatar.setStyle(mAvatarStyle);
 
+            viewHolder.mAvatarView
+                    .init(new AvatarViewModel(mImageCacheWrapper), new IdentityNameFormatterImpl())
```
sriamar: This means that AvatarViewModel's life is controlled by AvatarView. It shouldn't be. It should be tied to life of this class. So, make AvatarViewModel a member of this class now, and pass in the reference to AvatarView. When we do the work on other UI items to separate them, things will become more clear.

layer-atlas/src/main/java/com/layer/ui/typingindicators/AvatarTypingIndicatorFactory.java
```diff
@@ -91,39 +92,40 @@ public void onBindView(LinearLayout l, Map<Identity, LayerTypingIndicatorListene
...
 
         // Add new typists
         for (Identity typist : newlyActives) {
-            Avatar avatar = tag.mPassives.poll();
-            if (avatar == null) {
+            AvatarView avatarView = tag.mPassives.poll();
+            if (avatarView == null) {
                 // TODO: allow styling
-                avatar = new Avatar(l.getContext()).init(mPicasso);
+                avatarView = new AvatarView(l.getContext()).init(new AvatarViewModel(mImageCacheWrapper), new IdentityNameFormatterImpl());
```
sriamar: This means that AvatarViewModel's life is controlled by AvatarView. It shouldn't be. It should be tied to life of this class (TypingIndicatorFactory). So, make AvatarViewModel a member of this class now, and pass in the reference to AvatarView. When we do the work on other UI items to separate them, things will become more clear.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.layer.sdk.messaging.Identity;
+import com.layer.ui.util.imagecache.BitmapWrapper;
+import com.layer.ui.util.imagecache.ImageCacheWrapper;
+
+import java.lang.ref.WeakReference;
+
+/**
+ * Avatar interface exposes the interaction between the AvatarView and AvatarViewModel Avatar.
+ * @see Avatar.ViewModel is implemented by
```
sriamar: Is this comment not complete?

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,60 @@
...
+         */
+        ImageCacheWrapper getImageCacheWrapper();
+
+        /**
+         * Returns the initial base on the
+         * @see IdentityNameFormatter passed into the ViewModel
+         * @param identity
+         * @return
+         */
+
```
sriamar: Remove this space below method comment?

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public void fetchBitmap(BitmapWrapper bitmapWrapper) {
+
+        mImageCacheWrapper.fetchBitmap(bitmapWrapper, new ImageCacheWrapper.Callback() {
+            @Override
+            public void onSuccess() {
+                updateView();
+
```
sriamar: Remove extra line?

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
andela-wakinsanmi: @sriamar It is

 * Avatar interface exposes the interaction between the AvatarView and AvatarViewModel Avatar.
 * @see Avatar.ViewModel is implemented by (@see AvatarViewModel) 
 * @see AvatarView

andela-wakinsanmi: I will update to  @link, thanks

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
sriamar: Talked to Wale. He is confused about this as this was MessagesAdapter. Let's leave this as is. We can fix this things when we go to fixing MessagesAdapter to follow the new pattern.

layer-atlas/src/main/java/com/layer/ui/util/imagecache/PicassoImageCacheWrapper.java
sriamar: That is good then! Add a comment please.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,55 @@
...
+         * @see AvatarView#init(ViewModel, IdentityNameFormatter)
+         * @param view
+         */
+        void setView(WeakReference<View> view);
+
+        /**
+         * Fetch BitMap from ImageCacheWrapper Implentation passed into the ViewModel
+         * @param bitmapWrapper
+         */
+        void fetchBitmap(BitmapWrapper bitmapWrapper);
```
sriamar: Remove fetchBitmap() from here. Instead, do `viewModel.getImageCacheWrapper().fetchBitmap()` (similar to cancelBitmap()`

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -284,18 +283,20 @@ private boolean setClusterSizes() {
...
-                        targetUrl = null;
+                    if (url != null && url.trim().length() == 0) {
+                        url = null;
+                    }
+
+                    if (url != null) {
+                        bitmapWrapper.setWidth(size);
+                        bitmapWrapper.setHeight(size);
+                        bitmapWrapper.setMultiTransform(avatarCount > 1);
+                        mViewModel.fetchBitmap(bitmapWrapper);
```
sriamar: Do `mViewModel.getImageCacheWrapper().fetchBitmap()`.

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModel.java
```diff
@@ -0,0 +1,73 @@
...
+    public void setIdentityNameFormatter(IdentityNameFormatter identityNameFormatter) {
+        mIdentityNameFormatter = identityNameFormatter;
+    }
+
+    public String getInitialsForAvatarView(Identity added) {
+        return mIdentityNameFormatter.getInitials(added);
+    }
+
+    @Override
+    public void fetchBitmap(BitmapWrapper bitmapWrapper) {
```
sriamar: Implement this in View, and remove the updateView method.

layer-atlas/src/main/java/com/layer/ui/util/imagecache/PicassoImageCacheWrapper.java
```diff
@@ -0,0 +1,76 @@
...
+            @Override
+            public void onBitmapLoaded(Bitmap bitmap, Picasso.LoadedFrom from) {
+                bitmapWrapper.setBitmap(bitmap);
+                callback.onSuccess();
+                mTargets.remove(this);
+            }
+
+            @Override
+            public void onBitmapFailed(Drawable errorDrawable) {
+                bitmapWrapper.setBitmap(null);
```
sriamar: Add a log statement on failure. It could be verbose for now, as we will see failures in case of network issues.
if (Log.isLoggable(VERBOSE)) {
    Log.v("onBitMapFailed :" + errorDrawable);
}

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
andela-wakinsanmi: Hello @sriamar 

Thanks for the feedback, I removed this in my previous commit. The reason I brought it back is because the fetchBitmap in the ImageCacheWrapper takes in fetchBitmap(BitmapWrapper bitmapWrapper, Callback callback). Is it better for me to implement this callback in the AvatarView or in the AvatarViewModel?

sriamar: This callback only refreshes view. So, it is okay for that to be in AvatarView.

layer-atlas/src/main/java/com/layer/ui/avatar/Avatar.java
```diff
@@ -0,0 +1,48 @@
...
+import com.layer.sdk.messaging.Identity;
+import com.layer.ui.util.imagecache.ImageCacheWrapper;
+
+import java.lang.ref.WeakReference;
+
+/**
+ * @see Avatar.ViewModel exposes methods in the AvatarViewModel which are called by the {@link AvatarView}
+ * @see Avatar.ViewModel is implemented by {@link AvatarViewModel}
+ **/
+public interface Avatar {
```
smpete: Do we even need this wrapping interface anymore now that the View extends from android.view.View?

andela-wakinsanmi: Thanks @smpete Is it ok if I just call the interface AvatarViewModel instead?

thecombatwombat: I agree on the removal of the outer interface, but if we do that, what do we call the implementation @andela-wakinsanmi ?

layer-atlas/src/main/java/com/layer/ui/util/imagecache/BitmapWrapper.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This class wraps everything needed in a Bitmap
+ * The Bitmap is used in {@link com.layer.ui.avatar.Avatar.ViewModel#fetchBitmap(BitmapWrapper)}
+ * and also in {@link PicassoImageCacheWrapper}
+ * Properties can be added to BitmapWrapper to suit other ImageCache library
+ * We use Picasso in the implementation
+ */
+public class BitmapWrapper {
+
+    private final UUID mUniqueId;
```
thecombatwombat: IMO this can just be a field called `mId`, with `getID` method. It having a type `UUID` implies that it is unique anyway

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageCacheWrapper.java
```diff
@@ -0,0 +1,34 @@
+package com.layer.ui.util.imagecache;
+
+import android.graphics.Bitmap;
+
+public interface ImageCacheWrapper {
```
thecombatwombat: why remove the ImageView method here? Is the idea that we'll implement it when we need it?

andela-wakinsanmi: Yes @thecombatwombat

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -198,15 +201,15 @@ public void setShouldShowAvatarInOneOnOneConversations(boolean shouldShowAvatarI
...
+     * @return If the AvatarViewModel for the other participant in a one on one conversation will be shown
      * or not. Defaults to `true`.
      */
     public boolean getShouldShowAvatarPresence() {
         return mShouldShowAvatarPresence;
     }
 
     /**
-     * @param shouldShowPresence Whether the Avatar for the other participant in a one on one
+     * @param shouldShowPresence Whether the AvatarViewModel for the other participant in a one on one
```
thecombatwombat: Fix these comments. An Avatar (or AvatarView) is shown. An AvatarViewModel is not shown

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -67,13 +63,11 @@
         PAINT_BITMAP.setAntiAlias(true);
     }
 
-    private Picasso mPicasso;
     private Set<Identity> mParticipants = new LinkedHashSet<>();
 
-    // Initials and Picasso image targets by user ID
-    private final Map<Identity, ImageTarget> mImageTargets = new HashMap<>();
+    private final Map<Identity, BitmapWrapper> mIdentityBitmapWrapperHashMap = new HashMap<>();
```
thecombatwombat: The reference is of type `Map`, so the field should be called `mIdentityBitmapWrapperMap`

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -392,60 +419,10 @@ private void drawPresence(Canvas canvas, Identity identity) {
...
-            return mId == target.mId;
-        }
-
-        @Override
-        public int hashCode() {
-            return (int) (mId ^ (mId >>> 32));
-        }
+    private void parseStyle(Context context, AttributeSet attrs, int defStyleAttr) {
+        TypedArray ta = context.getTheme().obtainStyledAttributes(attrs, R.styleable.AvatarView, R.attr.AvatarView, defStyleAttr);
+        mMaxAvatar = ta.getInt(R.styleable.AvatarView_maximumAvatar, 3);
```
thecombatwombat: This property should be called `maximumAvatars`

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarViewModelImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public AvatarViewModelImpl(ImageCacheWrapper imageCacheWrapper) {
+        mImageCacheWrapper = imageCacheWrapper;
+    }
+
+    public void setIdentityNameFormatter(IdentityNameFormatter identityNameFormatter) {
+        mIdentityNameFormatter = identityNameFormatter;
+    }
+
+    public String getInitialsForAvatarView(Identity added) {
```
thecombatwombat: the parameter name here does not make sense. it should be 

`public String getInitialsForAvatarView(Identity identity) `

thecombatwombat: As it is in the interface....

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
thecombatwombat: that change will need to be in the xml as well I think?

PR #230, opened 2017-07-17
Decouple Presence from Avatar View

andela-wakinsanmi:


layer-atlas/src/androidTest/java/com/layer/ui/test/AvatarViewTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Test
+    public void testThatAvatarColorChangeWhenSpinnerIsChanged() {
+        String selectionText = "AWAY";
+
+        onView(withId(R.id.test_spinner)).perform(click());
+        onData(allOf(is(instanceOf(String.class)), is(selectionText))).perform(click());
+        onView(withId(R.id.test_spinner)).check(matches(withSpinnerText(containsString(selectionText))));
+        onView(withText(selectionText)).perform(click());
+        //onView(withId(R.id.test_presence)).check(matches(withBgColor(mAwayColor)));
```
thecombatwombat: remove if not needed

circle.yml
```diff
@@ -9,10 +9,15 @@ test:
         parallel: true
     # wait for it to have booted
     - circle-android wait-for-boot
+
+# Unlock the emulator screen after 15 seconds
+    - sleep 45
```
sriamar: that is 45 sec?

layer-atlas/src/main/java/com/layer/ui/presence/PresenceView.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Collection;
+import java.util.Set;
+
+public class PresenceView extends View {
+
+    private final Paint mBackgroundPaint = new Paint();
+    private final Paint mPresencePaint = new Paint();
+
+    private Identity mIdentity;
+    private int mAvailableColor = Color.rgb(0x4F, 0xBF, 0x62);
```
sriamar: All these colors should be pulled in from theme file?

layer-atlas/src/main/java/com/layer/ui/presence/PresenceView.java
```diff
@@ -0,0 +1,177 @@
...
+                drawOffline(canvas);
+                break;
+            case INVISIBLE:
+                drawInvisible(canvas);
+                break;
+            case BUSY:
+                drawBusy(canvas);
+                break;
+            default:
+                drawDefault(canvas);
```
sriamar: I don't understand this drawDefault. This switch statement is different from the one before in AvatarView().

layer-atlas/src/main/java/com/layer/ui/presence/PresenceView.java
```diff
@@ -0,0 +1,177 @@
...
+        canvas.drawCircle(presenceCenterX, presenceCenterY, presenceInnerRadius, mPresencePaint);
+
+        // Draw hollow if needed
+        if (makeCircleHollow) {
+            canvas.drawCircle(presenceCenterX, presenceCenterY, (presenceInnerRadius / 2f), mBackgroundPaint);
+        }
+    }
+
+    @VisibleForTesting
+    public int getPresenceColor() {
```
sriamar: This should be package private? i.e., leave out `public` ?

layer-atlas/src/main/java/com/layer/ui/presence/PresenceView.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public void setParticipants(Set<Identity> participants) {
+        showPresenceIfIdentityIsOne(participants);
+    }
+
+    public void setParticipants(Identity... participants) {
+        showPresenceIfIdentityIsOne(Arrays.asList(participants));
+    }
+
+    private void showPresenceIfIdentityIsOne(Collection<Identity> participants) {
```
sriamar: This method is not factored right. Here is all the things it does.
- if participant size is 1, it stores participant value in mIdentity. 
- Shows / hides presence
- requests redraw.
The name doesn't convey that. 


You can make the method name generic like `processSetParticipants()` instead?

sriamar:
And, your test failed! Let's fix that too :)

PR #238, opened 2017-08-01
Task/and 996 conversation item list layout

thecombatwombat:
Sizes conversation item based on item height specified by parent view. The Presence indicator does not work in the smallest/tiny size class, but that is a bug in Presence. Filed `AND-1042` to address that. See attached screenshots in Messenger

screenshot_1501552591

screenshot_1501552908

screenshot_1501552944

screenshot_1501552964


layer-atlas/src/main/java/com/layer/ui/style/ConversationItemStyle.java
```diff
@@ -1,74 +1,159 @@
...
+
+    // Cell background
+    private int mCellBackgroundColor;
+    private int mCellUnreadBackgroundColor;
+
+    // Margins
+    private float mMarginHorizontal;
+    private float mMarginVertical;
 
     public ConversationItemStyle(Context context, AttributeSet attrs, int defStyle) {
```
smpete: Maybe break this up into multiple private methods so it's easier to understand?

layer-atlas/src/main/java/com/layer/ui/style/ConversationItemStyle.java
```diff
@@ -1,74 +1,159 @@
...
-        avatarStyleBuilder.avatarBackgroundColor(ta.getColor(R.styleable.ConversationItemsListView_avatarBackgroundColor, context.getResources().getColor(R.color.layer_ui_avatar_background)));
-        avatarStyleBuilder.avatarBorderColor(ta.getColor(R.styleable.ConversationItemsListView_avatarBorderColor, context.getResources().getColor(R.color.layer_ui_avatar_border)));
+        avatarStyleBuilder.avatarTextColor(ta.getColor(R.styleable.ConversationItemsListView_avatarTextColor, resources.getColor(R.color.layer_ui_avatar_text)));
+        avatarStyleBuilder.avatarBackgroundColor(ta.getColor(R.styleable.ConversationItemsListView_avatarBackgroundColor, resources.getColor(R.color.layer_ui_avatar_background)));
+        avatarStyleBuilder.avatarBorderColor(ta.getColor(R.styleable.ConversationItemsListView_avatarBorderColor, resources.getColor(R.color.layer_ui_avatar_border)));
+        mAvatarVisibility = View.VISIBLE;
+
+        float heightMedium = resources.getDimension(R.dimen.layer_ui_item_height_medium);
```
smpete: Are we sure we want to do this based on item height? Or should we just have a style for the normal/medium/small/tiny layouts?

layer-atlas/src/main/java/com/layer/ui/util/ViewUtils.java
```diff
@@ -0,0 +1,12 @@
+package com.layer.ui.util;
+
+import android.content.Context;
+import android.util.DisplayMetrics;
+import android.util.TypedValue;
+
+public class ViewUtils {
+    public static float dipToPixels(Context context, float dipValue) {
```
smpete: I don't see the use of this anywhere. Do convenience methods like this make code easier or harder to read and understand?

thecombatwombat: Sorry... I thought I would need that, and then forgot to remove it. 
I'll get rid of this one. 

More generally speaking, I'm opposed to putting business logic in `Util` type classes, the way we have it now. But for repeated, tedious code, I think classes like this are useful. As long as we don't make one giant Utils class that does everything.

layer-atlas/src/main/java/com/layer/ui/style/ConversationItemStyle.java
thecombatwombat: We do want to do this based on item height. Since the item height is specified via the top level parent. 

We *really* need a better way to do styles, and that is a task I'd like to approach separately.

layer-atlas/src/main/java/com/layer/ui/util/ViewUtils.java
smpete: I'll agree with ^^

layer-atlas/src/main/java/com/layer/ui/style/ConversationItemStyle.java
thecombatwombat: @smpete I see your point. I'd like to do this in the next set of things, i.e. `IdentityItem` and `IdentityList`. At that point, I should hopefully be able to pull some things into the base class and start identifying how to better organize these style objects.

PR #242, opened 2017-08-03
IdentityItem and IdentityItemList

thecombatwombat:
This PR creates a `FourPartItem` set of classes that covers the commonalities between `ConversationListItem` and `IdentityListItem`. So apologies for the bloat in PR size. 

I'm open to feedback on the naming of the `FourPartItem` stuff


sriamar:
Tests? Am I missing them?

thecombatwombat:
@sriamar ah no... I am. Lemme fix that

sriamar:
Add tests. Otherwise, things looks good.

layer-atlas/src/main/java/com/layer/ui/identity/IdentityFormatterImpl.java
```diff
@@ -0,0 +1,64 @@
+package com.layer.ui.identity;
+
+import android.text.TextUtils;
+
+import com.layer.sdk.messaging.Identity;
+
+public class IdentityFormatterImpl implements IdentityFormatter {
```
andela-wakinsanmi: Will it be better if we move this implementation to Layer Messenger?

andela-wakinsanmi: Will it be better if we move this implementation to Layer Messenger?

thecombatwombat: I've been thinking about it... and I think we should provide a default implementation, and then provide an alternative on in messenger to demonstrate the use case

andela-wakinsanmi: @thecombatwombat That makes sense, I think the IdentityFormatterImpl should be the default Implementation that resides in Messenger.  We only need to use interface in Atlas-Android, when client clones Atlas-Android, they can provide their own implementation. What do you think?

jwyld: Remember the ultimate goal is to make setup and use a matter of 3 steps:
1. Allocation & initialization
2. Set the model
3. Display

We should always try to give the developer a sensible default.  This gets them to a working solution in the fastest manner possible.  As they decide to customize, they can do that using the API we expose.

thecombatwombat: @andela-wakinsanmi In all the key places that it matters, we are using only the interface. The `*Impl` classes we're talking about, should live in LayerUI since we aim to maximize simplicity as Jeremy mentions above. 

If a client has no interest in doing a custom IdentityFormatter, then we have no reason to force them into it. 

That being said, the current viewmodels _do_ require some of these formatters as dependencies during initialization.  for example: 
```
public ConversationItemsListViewModel(Context context, LayerClient layerClient,
                                            ConversationItemFormatter conversationItemFormatter,
                                            ImageCacheWrapper imageCacheWrapper,
                                           IdentityFormatter identityFormatter) {
```

@jwyld Are you saying you'd prefer a 

```
public ConversationItemsListViewModel(Context context, LayerClient layerClient)
```
That automatically assumes default implementations for the other objects?  

I for one, am not particularly fond of that. I think the former strikes a balance between ease of use and having the developer discover the API and ways they can customize these components, without having to override or replace anything.

thecombatwombat:
I would also like to note here that I abhor the `*Impl` naming convention for what are publicly visible, default implementations of certain abstract classes or interfaces. 

I much calling them what they are, default implementations of their interfaces or superclasses. 
i.e. `DefaultIdentityFormatter`, `DefaultConversationFormatter` etc. 

In case any of you have opinions about this right away, lets discuss it. Else we should talk sometime down the line about renaming these classes

PR #243, opened 2017-08-07
Feature/and 997/update message list to layer ui

andela-wakinsanmi:


layer-atlas/src/main/res/layout-v21/ui_message_item_cell_image.xml
```diff
@@ -1,22 +1,26 @@
 <?xml version="1.0" encoding="utf-8"?>
-<merge
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_width="wrap_content"
-    android:layout_height="wrap_content">
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <ImageView
-        android:id="@+id/cell_image"
+    <FrameLayout
```
andela-wakinsanmi: FrameLayout should be better optimize not to create unnecessary View, fix.

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -91,21 +89,17 @@
...
 
     private int readCount = 0;
     private boolean delivered = false;
     private Map<Identity, Message.RecipientStatus> statuses;
-    private LayerDateFormatter mLayerDateFormatter;
+    private DateFormatter mDateFormatter;
 
-    public MessagesAdapter(Context context, LayerClient layerClient, ImageCacheWrapper imageCacheWrapper, LayerDateFormatter layerDateFormatter) {
+    public MessagesAdapter(Context context, LayerClient layerClient, ImageCacheWrapper imageCacheWrapper, DateFormatter dateFormatter) {
+        super(context, layerClient, "", false);
```
thecombatwombat: Use a meaningful value for `tag` in this constructor

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -271,7 +256,7 @@ public MessagesAdapter setOnMessageAppendListener(OnMessageAppendListener listen
      */
     public MessagesAdapter addCellFactories(CellFactory... cellFactories) {
         for (CellFactory cellFactory : cellFactories) {
-            cellFactory.setStyle(mMessageStyle);
+            cellFactory.setStyle(mStyle);
```
thecombatwombat: In cases like this, I'd prefer we use `getStyle()` rather than the superclass' member variable directly. 

Thoughts?

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -330,15 +311,15 @@ public ViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
...
         if (mFooterView != null && position == mFooterPosition) {
             // Footer
-            bindFooter(viewHolder);
+            bindFooter(holder);
         } else {
             // Cell
-            bindCellViewHolder((CellViewHolder) viewHolder, position);
+            bindCellViewHolder((CellViewHolder) holder, position);
         }
+        super.onBindViewHolder(holder, position, payloads);
```
thecombatwombat: This is calling _both_local methods _and_ the superclass `onBindViewHolder` ? why?

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -717,15 +687,19 @@ public static ClusterType fromMessages(Message older, Message newer) {
         void onMessageAppend(MessagesAdapter adapter, Message message);
     }
 
-    static class ViewHolder extends RecyclerView.ViewHolder {
-        public final static int RESOURCE_ID_FOOTER = R.layout.ui_message_item_footer;
+    static class ViewHolder extends ItemViewHolder<Message, MessageItemViewModel, ViewDataBinding, MessageStyle> {
+        //public final static int RESOURCE_ID_FOOTER = R.layout.ui_message_item_footer;
```
thecombatwombat: If it is not needed, please delete it.

layer-atlas/src/main/java/com/layer/ui/messageitem/MessageItemViewModel.java
```diff
@@ -24,23 +25,28 @@
     private String mRecipientStatus;
     private String mGroupTime;
     private boolean mRecipientStatusVisible;
+    private boolean mIsMyCellType;
 
     public boolean isClusterSpaceVisible() {
         return mIsClusterSpaceVisible;
     }
 
     public int isDisplayAvatar() {
```
thecombatwombat: should be called `shouldDisplayAvatar` or `isAvatarDisplayed`

layer-atlas/src/main/res/layout/ui_message_item.xml
```diff
@@ -132,17 +134,34 @@
 
                 </FrameLayout>
 
+                <android.widget.Space
```
thecombatwombat: I would, at some near future date, like to rebuild these layouts to be flatter, using `ConstraintLayout` or something. We can leave it be for now.

layer-atlas/src/main/res/layout-v21/ui_message_item_cell_image.xml
thecombatwombat: Its okay for now. We can revisit all these layouts in a separate PR

layer-atlas/src/main/res/layout/ui_message_item_footer.xml
```diff
@@ -13,13 +13,15 @@
...
 -->
-<FrameLayout
-    android:id="@+id/swipeable"
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:paddingLeft="@dimen/layer_ui_padding_normal"
-    android:paddingRight="@dimen/layer_ui_padding_normal"
-    android:paddingTop="@dimen/layer_ui_message_cluster_space">
+<layout
```
thecombatwombat: Why have an outer `layout` tag if no data binding is being used here?

layer-atlas/src/main/res/layout/ui_message_item_cell_image.xml
```diff
@@ -1,22 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
-<merge
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_width="wrap_content"
-    android:layout_height="wrap_content">
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
```
thecombatwombat: Why have an outer `layout` tag if no data binding is being used here?

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageWrapper.java
```diff
@@ -0,0 +1,162 @@
+package com.layer.ui.util.imagecache;
+
+import android.net.Uri;
+import android.support.v4.widget.ContentLoadingProgressBar;
+import android.widget.ImageView;
+
+import java.lang.ref.WeakReference;
+
+
+public class ImageWrapper implements ImageCacheWrapper.Callback {
```
thecombatwombat: Not entirely convinced that this class needs to exist. Why can't we expose a method with an `ImageView` parameter in the standard `ImageCachWrapper`?

layer-atlas/src/main/java/com/layer/ui/messagetypes/threepartimage/ThreePartImageCellFactory.java
```diff
@@ -74,7 +71,13 @@ public boolean isBindable(Message message) {
 
     @Override
     public CellHolder createCellHolder(ViewGroup cellView, boolean isMe, LayoutInflater layoutInflater) {
-        return new CellHolder(layoutInflater.inflate(R.layout.ui_message_item_cell_image, cellView, true));
+        ImageWrapper imageWrapper = new ImageWrapper.Builder()
```
thecombatwombat: See comment about `ImageWrapper` class

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -65,11 +65,10 @@
  *
  * @see CellFactory
  */
-public class MessagesAdapter extends RecyclerView.Adapter<MessagesAdapter.ViewHolder> implements
-        BaseAdapter<Message>, RecyclerViewController.Callback {
+public class MessagesAdapter extends ItemRecyclerViewAdapter<Message, MessageItemViewModel,
```
thecombatwombat: I strongly believe this class should be broken down into smaller classes. There is too much logic all wrapped up in here.

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -717,15 +687,19 @@ public static ClusterType fromMessages(Message older, Message newer) {
...
 
         // View cache
         protected ViewGroup mRoot;
 
-        public ViewHolder(View itemView) {
-            super(itemView);
-            mRoot = (ViewGroup) itemView.findViewById(R.id.swipeable);
+        public ViewHolder(ViewDataBinding viewDataBinding, MessageItemViewModel messageItemViewModel) {
+            super(viewDataBinding, messageItemViewModel);
+            if (viewDataBinding instanceof UiMessageItemBinding) {
```
thecombatwombat: Why can't these decisions be made in the `onCreateViewHolder` method?

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
```diff
@@ -736,88 +710,63 @@ public ViewHolder(View itemView) {
         // Cell
         protected CellFactory.CellHolder mCellHolder;
         protected CellFactory.CellHolderSpecs mCellHolderSpecs;
-        protected ViewDataBinding mViewDataBinding;
-        private MessageItemViewModel mMessageItemViewModel;
         private DateFormat mTimeFormat;
 
 
-        public CellViewHolder(ViewDataBinding viewDataBinding, MessageItemViewModel messageItemViewModel,
+        public CellViewHolder(UiMessageItemBinding uiMessageItemBinding, MessageItemViewModel messageItemViewModel,
```
thecombatwombat: Why not have 2 separate classes, `MessageItemViewHolder` and `MessageItemFooterViewHolder` and have them both inherit from `ItemViewHolder` ?

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
andela-wakinsanmi: I will go ahead and use the getter method, since the definition is protected, it can be accessed by the subclass, if we want to enforce the subclass to use getter, do we make the declaration private instead?

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
andela-wakinsanmi: @thecombatwombat The super onBindViewHolder method seems to be doing clean up task like updating the Bound Position and executing Pending Bindings. I though I should delegate the task to the super.onBindViewHolder. Do you have feedback on a better way to make this cleaner?

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
andela-wakinsanmi: I will get rid it, thanks

layer-atlas/src/main/java/com/layer/ui/messageitem/MessageItemViewModel.java
andela-wakinsanmi: I will change that, thanks

layer-atlas/src/main/res/layout/ui_message_item_cell_image.xml
andela-wakinsanmi: I added that basically to get rid of the findViewById in the CellFactories that are inflating this layout. I don't see any reason to bind a ViewModel to it because the ImageView is being loaded by Picasso.

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageWrapper.java
andela-wakinsanmi: Exposing a method with ImageView and other parameters needed by Picasso to load the Image will be long. I decided to use the approach we used for BitmapWrapper for consistency sake and I think it's cleaner than passing lots of parameters in the method. A Glide user can simply go to the class and add the fields needed to make Glide work for them

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
andela-wakinsanmi: Lots of logic, thanks for the feedback. I will break things down here.

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
andela-wakinsanmi: This is to differentiate between the exact mRoot view to be used because the adapter inflates Ui_message_item_footer and ui_message_item.

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
andela-wakinsanmi: Thanks, I will do that

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
thecombatwombat: I don't particularly want to force anything here, its just a matter of what convention we use.

I would prefer that _we_ use the getters, but don't want to impose that on the developers.

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
thecombatwombat: I worry about things like `binding.executePendingBindings()` being called multiple times. 

My recommendation is that you verify that that is not happening, and if so, I'm fine with what you have, otherwise, override wholesale and perform all those tasks in the overridden method

layer-atlas/src/main/res/layout/ui_message_item_cell_image.xml
thecombatwombat: The `ImageView` is not being loaded by picasso... the imageview is part of the view/layout. 
Also, you can write a binding adapter that loads images using the `ImageCacheWrapper`. 

If adding the `layout` tag eliminates the `findViewById` calls, I'm okay with it

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageWrapper.java
thecombatwombat: 1) The `BitmapWrapper` was needed for state management around in-flight and cancelled requests. 

2) Why do you need to expose a bunch of parameters? Which of those parameters can you _assume_ for the instance of `ImageCacheWrapper` being used in `MessagesAdapter`? 
Can you simplify that by using a separate instance of `ImageCacheWrapper` ?

thecombatwombat: Also... the whole point of writing these wrappers is that no one should _have_ to go in and change any fields. So for someone using Glide, they can receive the same request, and then decide what action to take bases on the context they are receiving them in

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
thecombatwombat: You don't have to do anything very complex, but classes like `Cluster` and the `ViewHolder` s can be made independent

layer-atlas/src/main/java/com/layer/ui/adapters/MessagesAdapter.java
thecombatwombat: Fix the rest of the issues, and we'll come back and talk about this

layer-atlas/src/main/java/com/layer/ui/messagetypes/text/TextCellFactoryViewModel.java
```diff
@@ -0,0 +1,36 @@
+package com.layer.ui.messagetypes.text;
+
+import android.databinding.BaseObservable;
+import android.graphics.Typeface;
+
+import com.layer.ui.messagetypes.MessageStyle;
+
+public class TextCellFactoryViewModel extends BaseObservable {
```
thecombatwombat: This class is not needed. The style can be set on the layout binding directly, and used from there

layer-atlas/src/main/java/com/layer/ui/messageitem/MessageItemView.java
```diff
@@ -0,0 +1,163 @@
...
+        super(context, attrs);
+    }
+
+    public MessageItemView(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    public void init(LayoutInflater layoutInflater, ViewGroup parent, MessageItemViewModel messageItemViewModel, DateFormatter dateFormatter) {
+        mDateFormatter = dateFormatter;
+        mUiMessageItemBinding = UiMessageItemBinding.inflate(layoutInflater, parent, false);
```
thecombatwombat: You can also do `LayoutInflater.from(getContext()).inflate()`

layer-atlas/src/main/java/com/layer/ui/adapters/messages/MessageCellViewHolder.java
```diff
@@ -0,0 +1,62 @@
...
+import com.layer.ui.messageitem.MessageItemViewModel;
+import com.layer.ui.messagetypes.CellFactory;
+import com.layer.ui.messagetypes.MessageStyle;
+import com.layer.ui.util.imagecache.ImageCacheWrapper;
+
+public class MessageCellViewHolder extends
+        ItemViewHolder<Message, MessageItemViewModel, ViewDataBinding, MessageStyle> {
+
+    // Cell
+    protected CellFactory.CellHolder mCellHolder;
```
andela-wakinsanmi: That's the cellHolder in the CellFactory, CellHolders maintain a reference to their Message, and allow the capture of user interactions with their messages

layer-atlas/src/main/java/com/layer/ui/messagetypes/singlepartimage/SinglePartImageCellFactory.java
```diff
@@ -149,10 +150,14 @@ private Transformation getTransform(Context context) {
     public static class CellHolder extends CellFactory.CellHolder {
         ImageView mImageView;
         ContentLoadingProgressBar mProgressBar;
+        ViewDataBinding mViewDataBinding;
 
-        public CellHolder(View view) {
-            mImageView = (ImageView) view.findViewById(R.id.cell_image);
-            mProgressBar = (ContentLoadingProgressBar) view.findViewById(R.id.cell_progress);
+        public CellHolder(ViewDataBinding viewDataBinding) {
```
thecombatwombat: Why not directly accept an instance of `UiMessageItemCellBinding` ? The way it is done in `TextCellFactory` ?

layer-atlas/src/main/java/com/layer/ui/messagetypes/threepartimage/ThreePartImageCellFactory.java
```diff
@@ -275,12 +281,14 @@ public Info createFromParcel(Parcel in) {
...
     static class CellHolder extends CellFactory.CellHolder {
-        ImageView mImageView;
-        ContentLoadingProgressBar mProgressBar;
+        private ImageView mImageView;
+        private ContentLoadingProgressBar mProgressBar;
 
-        public CellHolder(View view) {
-            mImageView = (ImageView) view.findViewById(R.id.cell_image);
-            mProgressBar = (ContentLoadingProgressBar) view.findViewById(R.id.cell_progress);
+        public CellHolder(ViewDataBinding viewDataBinding) {
```
thecombatwombat: What happens if someone passes in the wrong viewdatabinding class in here?

layer-atlas/src/main/java/com/layer/ui/messagetypes/threepartimage/ThreePartImageCellFactory.java
```diff
@@ -275,12 +281,14 @@ public Info createFromParcel(Parcel in) {
...
+            if (viewDataBinding instanceof UiMessageItemCellImageBinding) {
```
thecombatwombat: Why not directly accept an instance of `UiMessageItemCellBinding` ? The way it is done in `TextCellFactory` ?

layer-atlas/src/androidTest/java/com/layer/ui/test/GenericCellFactoryTest.java
```diff
@@ -7,7 +7,7 @@
 import com.layer.sdk.LayerClient;
 import com.layer.sdk.messaging.Message;
 import com.layer.sdk.messaging.MessagePart;
-import com.layer.ui.messagetypes.generic.GenericCellFactory;
+import com.layer.ui.message.messagetypes.generic.GenericCellFactory;
```
thecombatwombat: I'm looking at the semantics of this, and I don't think you need a `messagetypes` package. All the things contained within that can be pulled up as is, one level.

andela-wakinsanmi: @thecombatwombat Do you mean I should remove the messagtypes to

com.layer.ui.message.generic.GenericCellFactory;
com.layer.ui.message.location.GenericCellFactory;
com.layer.ui.message.singlepart.GenericCellFactory;
com.layer.ui.message.text.GenericCellFactory;
com.layer.ui.message.threpartimage.GenericCellFactory;

instead?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemViewModel.java
```diff
@@ -148,4 +161,13 @@ public boolean isRecipientStatusVisible() {
     public void setRecipientStatusVisible(boolean recipientStatusVisible) {
         mRecipientStatusVisible = recipientStatusVisible;
     }
+
+    public void setMyCellType(boolean isMyCellType) {
```
thecombatwombat: More comments on boolean method and property naming:
`setIsDisplayName` -> `setShouldShowDisplayName`
`isDisplayName` -> `getShouldShowDisplayName`

`setIsBindDateTimeForMessage` -> `setShouldBindDateTimeForMessage`
`isBindDateTimeForMessage` -> `getShouldBindDateTimeForMessage`

`setClusterSpaceVisible` -> `setIsCluserSpaceVisible`
`setRecipientStatusVisible` -> `setIsRecipientStatusVisible`

layer-atlas/src/main/java/com/layer/ui/MessagesRecyclerView.java
```diff
@@ -59,32 +59,21 @@ public MessagesRecyclerView(Context context) {
         super(context);
     }
 
-    public MessagesRecyclerView init(LayerClient layerClient, ImageCacheWrapper imageCacheWrapper,
-            LayerDateFormatter layerDateFormatter) {
+    private MessagesRecyclerView init() {
```
andela-wakinsanmi: The init method will most likely have to remain because I need to initialize when the MessageRecyclerView is used in Java and xml. It will have a void return type

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
```diff
@@ -122,9 +122,11 @@ public void setParticipants(Identity... participants) {
      * Should be called from UI thread.
      */
     public void setParticipants(Set<Identity> participants) {
-        mParticipants.clear();
```
andela-wakinsanmi: @thecombatwombat The method has mParticipants.addAll(), which cannot take null as argument.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemViewModel.java
andela-wakinsanmi: The naming decision is because we are using JavaBeans object hence Databinding will be looking for isMethodName for predicate methods https://developer.android.com/topic/libraries/data-binding/index.html#data_object 

I will try with the getter and see if it works, if it doesnt, I can use the exact method name in the xm file, dataBinding should resolve that

layer-atlas/src/main/java/com/layer/ui/MessagesRecyclerView.java
```diff
@@ -57,40 +58,44 @@ public MessagesRecyclerView(Context context, AttributeSet attrs) {
...
 
-    public void setAdapter(ItemRecyclerViewAdapter adapter) {
+    private void throwExceptionIfNotMessageAdapter(Adapter adapter) {
         if (adapter == null || !(adapter instanceof MessagesAdapter)) {
-            throw new RuntimeException("AtlasMessagesRecyclerView sets MessagesAdapter");
+            throw new IllegalArgumentException("The argument passed is not of the correct type");
         }
+    }
+
+    public void setAdapter(ItemRecyclerViewAdapter adapter) {
```
thecombatwombat: Would this not be simpler if you just set the parameter type as `MessagesAdapter` ?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemViewModel.java
thecombatwombat: In databinding, `getShouldShowDisplayName` will resolve as `shouldShowDisplayName`

thecombatwombat: You can go ahead and make ☝️  changes, and they will work, as long as you update the layout files to use the correct names

layer-atlas/src/main/java/com/layer/ui/avatar/AvatarView.java
thecombatwombat: If you receive a null argument, throw a `NullPointerException` in case the underlying method won't accept it. Don't silently swallow the error.

PR #246, opened 2017-08-16
Convert MessageComposer to ComposeBar

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/composebar/ComposeBar.java
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    @Override
+    public void afterTextChanged(Editable editable) {
+        if (mConversation == null || mConversation.isDeleted()) return;
+        if (editable.length() > 0) {
+            if (!mSendButton.isEnabled()) {
+                mSendButton.setEnabled(true);
+            }
+            mConversation.send(LayerTypingIndicatorListener.TypingIndicator.STARTED);
```
smpete: I don't think logic like this should be in the View

thecombatwombat: Logic like sending Typing indicators or managing state of subviews?

smpete: Like sending typing indicators

thecombatwombat: Well it ports over existing behavior, and if you think of its intended use, it'll fall into 2 camps:

1) Use in `ConversationView`, which is a (more or less) drop in component where you want to abstract away all the complexity involved in displaying a conversation. In which case, that logic could live here or be part of another view, namely the `ConversationView`

2) Use as a standalone component, wherein I understand that there may be a case for not wanting to send Typing Indicators, and we should allow for that. 

But if we are looking to set the conversation on this object and having it send messages, then the logic for sending Typing Indicators correctly lives in `ComposeBar`. 

If we do take it out of here, how do you recommend we do so? A delegate for the containing view or activity? of perhaps the form:

```
public interface OnComposeBarTextStateListener {
     public void onSendButtonPressed(Editable editable);
     public void afterTextChanged(Editable editable);
}```

I don't care for ☝️  since we want a component that is as easy as possible to set up and use, and this forces some of that complexity on the developer. 

If a dev needs behavior different than what is implemented here, subclassing `ComposeBar` would be their best bet. The class specifically avoids having private members to allow for as much flexibility as possible when subclassing.

File: layer-atlas/src/main/res/values/ui-styles.xml [hunk 2]
```diff
@@ -112,4 +101,29 @@
...
+        <item name="inputTextSize">@dimen/layer_ui_text_size_input</item>
+        <item name="inputTextTypeface">normal</item>
+        <item name="inputTextStyle">normal</item>
+        <item name="android:enabled">true</item>
+        <item name="inputUnderlineColor">@color/layer_ui_color_primary_blue</item>
+        <item name="inputCursorColor">@color/layer_ui_color_primary_blue</item>
+        <item name="attachmentSendersBackground">@drawable/ui_popup_background</item>
+    </style>
+
+    <style name="ComposeBarImageButton" parent="@style/Widget.AppCompat.Button.Borderless">
```
smpete: Maybe make this a styleable attribute on `ComposeBar`?

layer-atlas/src/main/java/com/layer/ui/composebar/ComposeBar.java
smpete: I think this makes sense. 👍

layer-atlas/src/main/res/values/ui-styles.xml
thecombatwombat: *groan* turns out you can't set a `Style` programmatically on an existing view. 🤦‍♂️ 

https://stackoverflow.com/questions/11723881/android-set-view-style-programmatically

Doing that would involve parsing and manually applying everything there. Tedious. The ImageButtons on the left and right approach is going to be revisited before we ship anyway. 

@smpete How do you feel about punting on the item until then? How we approach those buttons is going to be up for discussion anyway

PR #247, opened 2017-08-21
WIP **** Feature/and 999/typing indicator to layer ui

andela-wakinsanmi:


layer-atlas/src/main/res/values/ui-strings.xml
```diff
@@ -59,4 +59,10 @@
     <string name="layer_ui_message_composer_hint">Message</string>
     <string name="layer_ui_message_composer_send_button">SEND</string>
 
+    <plurals name="layer_ui_typing_indicator_message">
+        <item quantity="one">%1$s, %2$s and %3$d person are typing</item>
```
thecombatwombat: how is this item quantity `one` ? There are three items here. Please name accordingly

layer-atlas/src/main/java/com/layer/ui/message/MessageItemFooterViewHolder.java
```diff
@@ -1,19 +1,68 @@
...
+
+        super(binding, messageItemViewModel);
         mRoot = binding.swipeable;
+        binding.avatar.init(new AvatarViewModelImpl(imageCacheWrapper), new IdentityFormatterImpl());
+    }
+
+    public void bind(Set<Identity> users, View mFooterView, int avatarViewVisibilityType) {
+
+        mRoot.addView(mFooterView);
+        mViewModel.setParticipants(users);
```
andela-wakinsanmi: Thanks, I will do that

layer-atlas/src/main/res/values/ui-strings.xml
andela-wakinsanmi: @thecombatwombat the one is meant to be used when we have one person left. It's just a way to toggle between the appropriate resource. According to https://developer.android.com/guide/topics/resources/string-resource.html#Plurals `The full set supported by Android is zero, one, two, few, many, and other.`

layer-atlas/src/main/java/com/layer/ui/message/MessageItemViewModel.java
```diff
@@ -24,6 +23,9 @@
     private boolean mIsRecipientStatusVisible;
     private boolean mIsMyCellType;
     private int mAvatarViewVisibilityType;
+    private int mMessageFooterAnimationVisibility;
```
andela-wakinsanmi: @thecombatwombat I can set a boolean on the ViewModel and use viewModel.getVisibilty ? View.VISIBLE : View:GONE for that state ?

thecombatwombat: I don't know... see a `setVisibility` method should allow for all valid visibility values... not sure if we should restrict it to a boolean, and therefore only 2 states. 

One way would be to define our own `enum` (which is what the perf guys at Android complain about so much) or if we can do some fancy annotation checks

smpete: Yep, I'd prefer a boolean to an int. INVISIBLE shouldn't ever be necessary.

layer-atlas/src/main/res/values/ui-strings.xml
thecombatwombat: I know its just a way to toggle between resources, but looking at the strings, I think the cases you have correctly map to `few` and `many`

layer-atlas/src/main/java/com/layer/ui/message/MessageItemViewModel.java
thecombatwombat: @andela-wakinsanmi ok... if you both think that a boolean is the right solution here, go ahead and do that. If needed, we can revisit this

layer-atlas/src/main/java/com/layer/ui/TypingIndicatorLayout.java
```diff
@@ -150,13 +151,18 @@ public void onTypingIndicator(LayerClient layerClient, Conversation conversation
         } else {
             mTypists.put(user, typingIndicator);
         }
+
         empty = mTypists.isEmpty();
```
thecombatwombat: rename `empty` to `isEmpty`

layer-atlas/src/main/res/layout/ui_message_item.xml
```diff
@@ -111,26 +111,36 @@
...
+
+
+                        <com.layer.ui.presence.PresenceView
+                            android:id="@+id/presence"
+                            android:layout_width="@dimen/layer_ui_message_avatar_item_single"
+                            android:layout_height="@dimen/layer_ui_message_avatar_item_single"
+                            android:layout_gravity="bottom|right" />
+
+                    </FrameLayout>
+                    <!--Empty space when Avatar Visibilty is meant to be Invisible so that messages can align well-->
```
thecombatwombat: Fix comment wrt:
1) Typo in Visibility
2) `is meant to be View.GONE`

PR #249, opened 2017-08-23
in Dubizzle, we have a requirement to:

hishamdubizzle:
in Dubizzle, we have a requirement to:
-  in conversations list screen: display participant name . make participant name bold if there is unread message
- in conversation messages screen: display participant name, photo, presence and last seen
- presence style change in case of online. presence should be hidden in case of other statuses

### DIFF
--- a/layer-atlas/src/main/java/com/layer/atlas/AtlasAvatar.java
+++ b/layer-atlas/src/main/java/com/layer/atlas/AtlasAvatar.java
@@ -349,6 +349,7 @@ public class AtlasAvatar extends View {
     }
 
     private void drawPresence(Canvas canvas, Identity identity) {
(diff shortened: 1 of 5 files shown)


thecombatwombat:
@hishamdubizzle Thanks for making a PR! 

One of the problems you seem to face is that Atlas is currently not flexible enough for your needs. We're hoping to change that with the next version of Atlas (called Layer UI). 

I'm going to take the changes you've done above as feedback for what we're doing right now, and hopefully Layer UI will let you achieve all of the above without having to make your own fork. 

As far as merging these changes is concerned, we cannot simply merge them in, since there are many other developers who depend on Atlas working a certain way, and your changes might not be consistent with what they need. We can't make all these changes for other developers without making them all optional. 

So for the time being, I think the best place for these changes is in your own fork.

PR #260, opened 2017-09-11
Task/and 596/ability to add default first view in message list

andela-wakinsanmi:


layer-atlas/src/main/java/com/layer/ui/message/MessageItemListView.java
```diff
@@ -73,6 +74,9 @@ public void onDestroy() {
 
     public void setAdapter(final MessagesAdapter adapter) {
         adapter.setStyle(mMessageStyle);
+        if (mHeaderView != -1) {
+            adapter.setHeaderView(mHeaderView);
```
thecombatwombat: Why does a method with the name `setHeaderView` take an integer as an argument?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemListView.java
```diff
@@ -168,13 +172,19 @@ private int findLastVisibleItemPosition() {
...
+            mAdapter.setParticipants(conversation.getParticipants());
         }
+
         mAdapter.setQuery(Query.builder(Message.class)
                 .predicate(new Predicate(Message.Property.CONVERSATION, Predicate.Operator.EQUAL_TO, conversation))
                 .sortDescriptor(new SortDescriptor(Message.Property.POSITION, SortDescriptor.Order.ASCENDING))
                 .build()).refresh();
     }
 
+    public void setHeaderView(int headerView) {
```
thecombatwombat: Why does a method with the name setHeaderView take an integer as an argument?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
```diff
@@ -5,22 +5,46 @@
...
+    protected SwipeableItem.OnItemSwipeListener<Message> mItemSwipeListener;
+    protected boolean mIsMessageSizeZero;
+    protected String emptyMessageText;
+    private LayerClient mLayerClient;
+    private EmptyMessageFormatter mEmptyMessageFormatter;
+
+    protected MessagesAdapter.MessageAdapterEmptyRegister mMessageAdapterEmptyRegister = new MessagesAdapter.MessageAdapterEmptyRegister() {
+        @Override
+        public void setIsMessageSizeZero(boolean isMessageSizeZero, Set<Identity> participants) {
+            mIsMessageSizeZero = isMessageSizeZero;
```
thecombatwombat: If the adapter is going to determine what happens when the number of messages is 0, then why is all this stuff here in the viewmodel?

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -82,19 +85,25 @@
...
     protected int mViewTypeCount = VIEW_TYPE_FOOTER;
     protected boolean mShouldShowAvatarInOneOnOneConversations;
+    protected boolean mShouldDisplayFirstViewInMessageList;
     protected boolean mShouldShowAvatarPresence = true;
     private View mFooterView;
     private int mFooterPosition = 0;
     private Integer mRecipientStatusPosition;
     private boolean mReadReceiptsEnabled = true;
     private ImageCacheWrapper mImageCacheWrapper;
+    private int mMessageItemCount = -1;
```
thecombatwombat: The adapter should not track message item count state. That is the job of `getItemCount()`

thecombatwombat: There can NEVER be -1 messages in a conversation.

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -293,15 +306,37 @@ public int getItemViewType(int position) {
     @Override
     public ItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
 
-        if (viewType == VIEW_TYPE_FOOTER) {
-            MessageItemViewModel messageItemViewModel = new MessageItemViewModel(null);
+        //Create first
+        if (mShouldDisplayFirstViewInMessageList && viewType == VIEW_TYPE_DEFAULT_FIRST_VIEW) {
```
thecombatwombat: `mShouldDisplayFirstViewInMessageList` 

Please use a better variable name. When would there ever not be a first view in a messages list? 

eg:
`mShouldDisplayEmptyListView`

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -279,8 +288,12 @@ public MessagesAdapter addCellFactories(List<CellFactory> cellFactories) {
 
     @Override
     public int getItemViewType(int position) {
+        //When user wants a default view as the first message
+        if (mShouldDisplayFirstViewInMessageList && position == 0) return VIEW_TYPE_DEFAULT_FIRST_VIEW;
         if (mFooterView != null && position == mFooterPosition) return VIEW_TYPE_FOOTER;
-        Message message = getItem(position);
+
+        int index = mShouldDisplayFirstViewInMessageList ? position - 1 : position;
+        Message message = getItem(index);
```
thecombatwombat: what happens when ` mShouldDisplayFirstViewInMessageList==true  && position==0`? 

`index = -1`. what happens to `getItem(position)` when you give it an argument `-1`

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -293,15 +306,37 @@ public int getItemViewType(int position) {
...
-            return new MessageItemFooterViewHolder(uiMessageItemFooterBinding, messageItemViewModel,
-                    mImageCacheWrapper);
+        if (viewType == VIEW_TYPE_FOOTER) {
+            return createFooterBindingView(parent);
         }
 
+        return createMessageItemBindingView(parent, viewType);
+    }
+
+    private ItemViewHolder createFirstViewInMessageList(ViewGroup parent) {
```
thecombatwombat: it is not the first view in a messages list. it is a view displayed when the list is empty, or a header view. 

please name the method as such

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -671,6 +731,21 @@ public void onQueryItemMoved(RecyclerViewController controller, int fromPosition
...
+    setMessageAdapterEmptyRegister(
+            MessageAdapterEmptyRegister messageAdapterEmptyRegister) {
+        mMessageAdapterEmptyRegister = messageAdapterEmptyRegister;
+    }
+
+    public void setParticipants(Set<Identity> participants) {
+        mParticipants = participants;
+    }
+
+    public void setHeaderView(int headerView) {
```
thecombatwombat: why does a method called `setHeaderView` take an argument of type `int` ?

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -684,4 +759,12 @@ public void onQueryItemMoved(RecyclerViewController controller, int fromPosition
          */
         void onMessageAppend(MessagesAdapter adapter, Message message);
     }
+
+    /**
+     * Call back use to notify the View when the Data source in the Adapter is empty
+     * The visibility of the Empty View can be toggled on and off
+     */
+    interface MessageAdapterEmptyRegister {
```
thecombatwombat: What is this? 

If the adapter needs to know if the message count is 0, it will call `getItemCount` and get back 0

layer-atlas/src/main/java/com/layer/ui/util/AvatarStyle.java
```diff
@@ -89,6 +96,11 @@ public Builder height (float height) {
             return this;
         }
 
+        public Builder presenceRadius(float presenceRadius) {
```
thecombatwombat: Why are these changes in this branch? If work on them has been done elsewhere, please rebase

layer-atlas/src/main/java/com/layer/ui/message/EmptyMessageFormatterImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.layer.sdk.messaging.Identity;
+
+import java.util.Set;
+
+public class EmptyMessageFormatterImpl implements EmptyMessageFormatter {
+
+    @Override
+    public String createEmptyMessageListText(Set<Identity> participants, Identity authenticatedUser) {
+        int counter = 1;
```
thecombatwombat: What does this method do? 

As far as I can tell, this can be accomplished by:

```
participants.remove(authenticatedUser);

if (participants.size() == 1 ) {
    emptyMessage = emptyMessage + " " + participants.get(0).getDisplayName();
} else if (participants.size() == 2) {
    emptyMessage = emptyMessage + " " + participants.get(0).getDisplayName() + " & " + participants.get(1).getDisplayName();
} else if (participants.size() == 3) {
    emptyMessage = emptyMessage + " " + participants.get(0).getDisplayName() + " & " + participants.get(1).getDisplayName() + " & " + "one other";
} else {
    emptyMessage = emptyMessage + " " + participants.get(0).getDisplayName() + " & " + participants.get(1).getDisplayName() + " & " + "others";
}

```

thecombatwombat: This can be done entirely without this method or this class, simply by using a plurals string

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
andela-wakinsanmi: Should the Adapter determine what to do when Message size is zero? The approach taken is for the activity to display the empty message when the Message is empty. Any idea on how to turn the Emtpy view on from the adapter, do I use a  broadcast?

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
andela-wakinsanmi: Yes, this is just to simply initialize it as against allowing the compiler to give it a default value

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
andela-wakinsanmi: The Header View is an XML Resource, I should have annotated it with @XmlRes.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemListView.java
andela-wakinsanmi: The Header View is an XML Resource, I should have annotated it with @xmlres.

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
andela-wakinsanmi: @thecombatwombat This is to display the HeaderView for the message, will mShouldDisplayHeaderView be appropriate?

PR #274, opened 2017-10-02
Refactor MessageAdapter + Card binding

thecombatwombat:
Some issues/open questions :

1. `ItemRecyclerViewAdapter` is not the perfect class to be used for a list of hetrogenous items. We're losing some type safety due to java's type erasure in MessageItemsAdapter. This is not a pressing concern at the moment, since `MessageItemsAdapter` is available as a default implementation of the API `MessagesAdapter` exposes, and most of the business logic is relegated to the CellFactories and card implementations.  We should seek to improve it before we ship. 

2. At some point, we should move out `MessageCluster` related logic to a separate class/component. This will allow us to switch clustering rules more freely, and be able to write tests for it. 

3. Should we be setting up some sensible defaults for most of these viewmodels? they seem to have really dependency heavy constructors

4. The `MessagesAdapter` could use some more reorganizing and tidying up.


layer-atlas/src/main/java/com/layer/ui/adapters/ItemViewHolder.java
```diff
@@ -14,25 +19,40 @@
...
 
     public ItemViewHolder(BINDING binding, VIEW_MODEL viewModel) {
         super(binding.getRoot());
         mBinding = binding;
         mViewModel = viewModel;
+        mLayoutInflater = LayoutInflater.from(binding.getRoot().getContext());
+    }
+
+    public ItemViewHolder(ViewGroup parent, @LayoutRes int layoutId, VIEW_MODEL viewModel) {
+        this((BINDING) DataBindingUtil.inflate(LayoutInflater.from(parent.getContext()), layoutId, parent, false), viewModel);
```
smpete: You can type this call to avoid the cast:

`
this(DataBindingUtil.<BINDING>inflate(LayoutInflater.from(parent.getContext()), layoutId, parent, false), viewModel);
`

smpete:
Agreed on moving `MessageCluster` logic out of the adapter. That'll clean things up.

I do think we should have sensible defaults for the view models. Single arg constructors should be possible for the common use case.

thecombatwombat:
@smpete I've provided defaults for `DateFormatter` and `IdentityFormatter` all the way up in `ItemViewModel<ITEM>`. A single arg constructor is not quite possible, since we need both `Context` and `LayerClient`. 

I much prefer that the `ImageCacheWrapper` implementations come from the developers. Any thoughts on this?

thecombatwombat:
@smpete I'm going to punt on moving the clustering stuff out of the `MessageAdapter` just yet. There's some message update logic baked into that right now, that I'd don't want to mess with just yet. 

But I'll add a task for us to get to it later on.

smpete:
@thecombatwombat That's likely good enough. We should be able to get rid of `Context` if we go to `android.arch.lifecycle.ViewModel`s.

We're still bundling Picasso with it right? If so then I'd think we should have the default just use the provided Picasso wrapper. If we are not, then it should be a constructor arg.

layer-atlas/src/main/java/com/layer/ui/viewmodel/ItemViewModel.java
```diff
@@ -1,18 +1,35 @@
...
 import com.layer.sdk.query.Queryable;
+import com.layer.ui.identity.IdentityFormatter;
+import com.layer.ui.identity.IdentityFormatterImpl;
 import com.layer.ui.recyclerview.OnItemClickListener;
+import com.layer.ui.util.DateFormatter;
+import com.layer.ui.util.DateFormatterImpl;
 
 public class ItemViewModel<ITEM extends Queryable> extends BaseObservable {
 
+    protected Context mContext;
```
smpete: What's the point of protected members that have public accessors/mutators?

thecombatwombat:
@smpete We are bundling Picasso, although I think we've heard a fair number of complaints from developers saying that they'd like to switch it out to xyz. 

We do require a little special code (the custom request handler) and such for Picasso to work, and I'd rather have developers discover that and make a conscious choice to use Picasso, or implement something else.

I'm not married to this, so can still do it the way you mentioned. Lemme know what you think.

PR #277, opened 2017-10-04
Fix crash related to Picasso resize when size is <= 0

andela-wakinsanmi:


layer-atlas/src/main/java/com/layer/atlas/messagetypes/singlepartimage/SinglePartImageCellFactory.java
```diff
@@ -61,9 +63,17 @@ public void bindCellHolder(final CellHolder cellHolder, PartId index, Message me
...
         cellHolder.mProgressBar.show();
-        mPicasso.load(index.mId).tag(PICASSO_TAG).placeholder(PLACEHOLDER)
-                .centerInside().resize(specs.maxWidth, specs.maxHeight).onlyScaleDown()
-                .transform(getTransform(cellHolder.mImageView.getContext())).into(cellHolder.mImageView, new Callback() {
+        RequestCreator requestCreator = mPicasso.load(index.mId).tag(PICASSO_TAG).placeholder(PLACEHOLDER)
+                .centerInside();
+        if (specs.maxWidth > 0 && specs.maxHeight > 0) {
+            requestCreator.resize(specs.maxWidth, specs.maxHeight);
+        } else if (Log.isLoggable(Log.ERROR)) {
+            Log.e("Width or Height of image passed into SinglePartImageCellFactory.bindCellHolder should be > 0");
```
smpete: Can we get consistency on this log? Here it states that the dimension should be > 0. In the others it just states that it is invalid.

PR #279, opened 2017-10-12
Display Empty Message when List is Empty

andela-wakinsanmi:
This displays message on a Textview when the MessageList is empty


layer-atlas/src/main/java/com/layer/ui/conversation/ConversationViewModel.java
```diff
@@ -23,12 +24,14 @@
     protected LayerClient mLayerClient;
     protected Query<Message> mQuery;
 
-    public ConversationViewModel(Context context, LayerClient layerClient, List<CellFactory> cellFactories,
+    public ConversationViewModel(Context context, LayerClient layerClient,
+                                 EmptyListMessageFormatter emptyListMessageFormatter,
```
smpete: I'd prefer to see this as just a view. There is no reason we should restrict users to use a text view here. We should still supply the default as a text view like you have it. 

I think this should also be a setter rather than a constructor argument. We already have too many of those...

layer-atlas/src/main/res/layout/ui_message_items_list.xml
```diff
@@ -1,10 +1,26 @@
 <?xml version="1.0" encoding="utf-8"?>
 <merge xmlns:android="http://schemas.android.com/apk/res/android"
-       android:layout_width="match_parent"
-       android:layout_height="match_parent">
+    xmlns:tool="http://schemas.android.com/tools">
 
-    <com.layer.ui.recyclerview.ItemsRecyclerView
-        android:id="@+id/ui_message_recycler"
+    <RelativeLayout
```
smpete: This can be a FrameLayout since both aren't displayed at the same time

layer-atlas/src/main/java/com/layer/ui/message/EmptyListMessageFormatterImpl.java
```diff
@@ -0,0 +1,32 @@
...
+import com.layer.sdk.messaging.Identity;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+public class EmptyListMessageFormatterImpl implements EmptyListMessageFormatter {
+
+    @Override
+    public String createEmptyMessageListText(Set<Identity> participants, Identity authenticatedUser) {
```
sriamar: There are lots of problems with this method.

Bugs:
1 - Strings are hard coded. It should be in resource file?
2 - What happens if there is one participant and that is the authenticated user?

Design & structure :
3 - Where is the UI Design from? Why not a generalized string "Start a Conversation!"
4 - If you do need to use this model, why not use a switch statement, as opposed to a whole bunch of `if else`. That would make reading, and structuring much easier.
5 - You are appending a bunch of strings. Why not use `StringBuilder` or use format for your `emptyMessage`?

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationViewModel.java
```diff
@@ -39,6 +43,10 @@ public void setConversation(Conversation conversation) {
         notifyChange();
     }
 
+    public void setEmptyListMessageFormatter(EmptyListMessageFormatter emptyListMessageFormatter) {
```
smpete: This should be something along the lines of `setEmptyView(View view)`. Similar to Android's `AdapterView`. We shouldn't need to define a new object to set an empty view.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,9 +47,40 @@
 
     protected LayerClient mLayerClient;
     protected Conversation mConversation;
+    private Set<Identity> mParticipants = new HashSet<>();
+    private TextView mEmptyListTextView;
+    private ViewGroup emptyMessagePlaceholder;
```
smpete: This can just be a View.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
```diff
@@ -46,7 +54,18 @@ public void setOnItemSwipeListener(SwipeableItem.OnItemSwipeListener<Message> on
...
     }
 
+    @Bindable
     public SwipeableItem.OnItemSwipeListener<Message> getItemSwipeListener() {
         return mItemSwipeListener;
     }
+
+    @Bindable
+    public boolean getIsMessageSizeZero() {
+        return mIsMessageSizeZero;
```
sriamar: Where is this value set?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -55,6 +91,9 @@ public MessageItemsListView(Context context, AttributeSet attrs) {
 
         inflate(getContext(), R.layout.ui_message_items_list, this);
         mMessagesRecyclerView = (ItemsRecyclerView<Message>) findViewById(R.id.ui_message_recycler);
+        emptyMessagePlaceholder = (FrameLayout) findViewById(R.id.emptyMessagePlaceholder);
```
smpete: No need to cast now

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
```diff
@@ -46,7 +54,18 @@ public void setOnItemSwipeListener(SwipeableItem.OnItemSwipeListener<Message> on
         notifyChange();
     }
 
+    @Bindable
     public SwipeableItem.OnItemSwipeListener<Message> getItemSwipeListener() {
         return mItemSwipeListener;
     }
+
+    @Bindable
+    public boolean getIsMessageSizeZero() {
```
sriamar: Where is this used?

layer-atlas/src/main/java/com/layer/ui/message/EmptyListMessageFormatterImpl.java
andela-wakinsanmi: Thanks @sriamar , I thought about putting the String in String plurals but the have different placeholders. I will place them as a separate string resource each.
oops!!! I didn't know that we can have a conversation with only the authenticated user in, I will cover the case.

"Start a Conversation" is better but this is the specification in MessageList on zeplin.

I will change it to switch statement and StringFormat. Thanks

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationViewModel.java
andela-wakinsanmi: @smpete this is not setting the EmptyView, Zeplin has 
screen shot 2017-10-13 at 12 26 28 am 

So I created the EmptyListMessageFormatter interface and I passed in a default implementation if the client doesn't provide his/her own implementation.

layer-atlas/src/main/java/com/layer/ui/message/EmptyListMessageFormatterImpl.java
andela-wakinsanmi: image

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
andela-wakinsanmi: @smpete It seems I have to pass in a ViewGroup to the inflater

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationViewModel.java
smpete: If our goal is extensibility without forking then being able to pass a pure `View` right course in my opinion. We would still provide a `TextView` that matches the comps in Zeplin.

layer-atlas/src/main/java/com/layer/ui/message/EmptyListMessageFormatterImpl.java
smpete: Any String we use that is displayed in the UI _must_ be pulled from `strings.xml`. This allows it to be localized.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
smpete: Ah touche

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationViewModel.java
thecombatwombat: I'm a little leery of having `ViewModel` s contain `View` s. The `setEmptyView` method should exist on the `ConversationView`

@andela-wakinsanmi yes, that is what the Zeplin design says, but if you check the confluence spec, it requires the ability to set a header view. the `Start a conversation with <names>` is just one kind of header view

smpete: I'd agree with keeping `View`s out of `ViewModel`s, especially if/when we switch to the architecture component's `ViewModel`

layer-atlas/src/main/java/com/layer/ui/message/EmptyListMessageFormatterImpl.java
thecombatwombat: This entire class and the corresponding interface, IMO, should not exist. All this logic should live in whatever custom view is being set as the header view. 

There is no requirement for us to be able to provide for custom messages in case the conversation has no messages in it. We have a requirement for being able to set a custom header view. The header view can be set and unset on whatever condition is deemed appropriate. The way typing indicators are handled by setting a footer view upon some condition/state arising, and unset on that state going away.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,8 +48,39 @@
...
+    private RecyclerView.AdapterDataObserver emptyObserver = new RecyclerView.AdapterDataObserver() {
+        @Override
+        public void onChanged() {
+            super.onChanged();
+            if (mAdapter != null && mAdapter.getItemCount() == 0 && mLayerClient != null
+                    && mEmptyView != null) {
+
+                emptyMessagePlaceholder.setVisibility(View.VISIBLE);
+
+                emptyTextView.setText(updateEmptyTextView(mParticipants, mLayerClient.getAuthenticatedUser()));
```
smpete: I don't think this empty view should be required to use our generated empty text. This class should determine when to show/hide the view but should not be responsible for setting the data inside the view. It should not determine the details of the empty view. 

The `EmptyMessageListView` would take either the conversation or participants and the text generation would happen then inside that view (or view model depending if we want to introduce that complexity for the empty view).

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationView.java
```diff
@@ -92,4 +94,8 @@ public void setTypingIndicator(TypingIndicatorLayout typingIndicator) {
     public void onDestroy() {
         mMessageItemListView.onDestroy();
     }
+
+    public void setEmptyListView(View emptyListView, @IdRes int emptyTextViewResId) {
```
smpete: I think this should just pass a view. It shouldn't be required to have a text view. I guarantee there will be empty views that just want to use an image or their own text.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
andela-wakinsanmi: Thanks @smpete 

To do that, the View has to implement an interface that we can call to set the Participants on it. Is it better to take this approach and we can cast the Interface to the View to be able to add the View to the ViewGroup?

thecombatwombat: @andela-wakinsanmi this case is practically identical to how the footer view works, and how the typing indicator is built by setting a view that receives participants in its view model

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -79,18 +109,40 @@ public void onRefresh() {
...
+            mAdapter.setHeaderViewViewModel(mMessageItemHeaderViewModel);
+        }
     }
 
     public void setAdapter(final MessagesAdapter adapter) {
-        adapter.setStyle(mMessageStyle);
         mAdapter = adapter;
+        mAdapter.setStyle(mMessageStyle);
+        if (mEmptyView != null) {
+            mAdapter.setShouldShowHeader(true);
```
smpete: Should this also set the header view ViewModel as is done when setting the empty view?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -249,21 +301,60 @@ public void setConversation(LayerClient layerClient, Conversation conversation)
...
         mLayerClient.registerEventListener(this);
-
         mAdapter.setQuery(query, null);
-        Set<Identity> participants = conversation.getParticipants();
-        if (participants != null) {
-            mAdapter.setIsOneOnOneConversation(conversation.getParticipants().size() == 2);
+        mAdapter.refresh();
+    }
+
+    protected String updateEmptyTextView(Set<Identity> participants, Identity authenticatedUser) {
```
smpete: This is no longer used here. It should be removed

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -90,6 +91,8 @@
     private DateFormatter mDateFormatter;
     private IdentityFormatter mIdentityFormatter;
     private Set<Identity> mUsersTyping;
+    private MessageItemHeaderViewModel mMessageItemHeaderViewModel;
+    protected Conversation mConversation;
```
smpete: This opens the variable up to modification outside this class. I'd prefer to see this as a private variable and accessed via a protected getter.

andela-wakinsanmi: I made that protected because it's only being used by the sub class (MessageItemsAdapter) to bind HeaderView. I will make it private and expose a getter.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
andela-wakinsanmi: At this point we are still using a View class, the HeaderView ViewModel is set in the Adapter, when the binding is happening. At this point we just have a view, when the emptyView is being added, it checks if it's an instance of the EmptyMessageListView and then sets the ViewModel, if it's just an ordinary View, it's added to the ViewGroup

smpete: Then is there a problem with the `setEmptyView` method? If that view does not use the `mMessageItemHeadViewModel` then should that not be set?

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationView.java
```diff
@@ -92,4 +95,13 @@ public void setTypingIndicator(TypingIndicatorLayout typingIndicator) {
     public void onDestroy() {
         mMessageItemListView.onDestroy();
     }
+
+
+    public void setEmptyListView(View emptyListView) {
```
thecombatwombat: Empty lists have a view on their headers, not special views for when they are empty. 

This should be simply `setHeaderView`

layer-atlas/src/main/java/com/layer/ui/conversation/ConversationViewModel.java
```diff
@@ -1,37 +1,23 @@
...
-
-import java.util.List;
 
 public class ConversationViewModel extends BaseObservable {
     protected Conversation mConversation;
     protected MessageItemsListViewModel mMessageItemsListViewModel;
     protected LayerClient mLayerClient;
     protected Query<Message> mQuery;
 
-    public ConversationViewModel(Context context, LayerClient layerClient, List<CellFactory> cellFactories,
```
thecombatwombat: I don't understand the purpose of this change. There is no reason to expose `MessageItemListViewModel` via `ConversationView`

thecombatwombat: Please revert this change

layer-atlas/src/main/java/com/layer/ui/message/MessageItemHeaderViewModel.java
```diff
@@ -0,0 +1,83 @@
...
+        mContext = context;
+        mLayerclient = layerClient;
+    }
+
+    public void setConversation(Conversation conversation) {
+        mConversation = conversation;
+    }
+
+    @Bindable
+    public String getEmptyMessage() {
```
thecombatwombat: The header view model is supposed to have no knowledge of anything called an "empty view". the view models know nothing about views at all. All of this is logic that should live in the view that will be set inside the header view

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,8 +43,38 @@
 
     protected LayerClient mLayerClient;
     protected Conversation mConversation;
+    private Set<Identity> mParticipants = new HashSet<>();
```
thecombatwombat: Why is `mParticipants` field needed when you have a `Conversation` object already available?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,8 +43,38 @@
...
+    private View mEmptyView;
```
thecombatwombat: This list should contain a `mHeaderView`,  not empty view

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,8 +43,38 @@
...
+    private MessageItemHeaderViewModel mMessageItemHeaderViewModel;
```
thecombatwombat: Item view models should not be part of the list view. This object does not belong here

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,8 +43,38 @@
...
+    private int mNumberOfItemsPerSync;
```
thecombatwombat: Why has `mNumberOfItemsPerSync` been introduced here?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -42,8 +43,38 @@
...
+    private boolean mShouldShowHeader;
```
thecombatwombat: Why has `mShouldShowHeader` been introduced here?

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -290,4 +343,8 @@ protected void parseStyle(Context context, AttributeSet attrs, int defStyle) {
         ta.recycle();
         mMessageStyle = messageStyleBuilder.build();
     }
+
+    public void setEmptyListViewModel(MessageItemHeaderViewModel messageItemHeaderViewModel) {
```
thecombatwombat: There is no reason to have this exposed here. And why does `setEmptyListViewModel` take a `MessageItemHeaderViewModel` ?

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -90,6 +91,8 @@
     private DateFormatter mDateFormatter;
     private IdentityFormatter mIdentityFormatter;
     private Set<Identity> mUsersTyping;
+    private MessageItemHeaderViewModel mMessageItemHeaderViewModel;
```
thecombatwombat: This should not be part of the adapter's state

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -168,10 +171,18 @@ public boolean getShouldShowHeader() {
         return mShouldShowHeader;
     }
 
-    public void setShouldShowHeader(boolean shouldShowHeader) {
+    protected void setShouldShowHeader(boolean shouldShowHeader) {
```
thecombatwombat: Why is this `protected` ?

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -168,10 +171,18 @@ public boolean getShouldShowHeader() {
...
         mShouldShowHeader = shouldShowHeader;
     }
 
+    public void setHeaderViewViewModel(MessageItemHeaderViewModel messageItemHeaderViewModel) {
```
thecombatwombat: Header view models should be generated and not set externally

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -341,7 +352,7 @@ public int getItemCount() {
         if (getQueryController() != null) {
             itemCount = getQueryController().getItemCount();
         } else {
-            itemCount = getItems().size();
+            itemCount = getItems() != null ? getItems().size() : 0;
```
thecombatwombat: 👍

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -573,6 +584,18 @@ public void onQueryItemMoved(RecyclerViewController controller, int fromPosition
         super.onQueryItemMoved(controller, fromPosition, toPosition);
     }
 
+    public void setConversation(Conversation conversation) {
+        mConversation = conversation;
+    }
+
+    public Conversation getConversation() {
```
thecombatwombat: `getConversation` is not needed

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -573,6 +584,18 @@ public void onQueryItemMoved(RecyclerViewController controller, int fromPosition
...
+        return mConversation;
+    }
+
+    public void setEmptyHeaderView(View emptyHeaderView) {
```
thecombatwombat: what happens when someone wants to set a non-empty header view?

thecombatwombat: There should only be a `headerView`

layer-atlas/src/main/res/layout/ui_empty_message_items_list.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+
```
thecombatwombat: why is this file named `ui_empty_message_items_list.xml` if there is no list inside it?

layer-atlas/src/main/res/layout/ui_message_item_card.xml
```diff
@@ -56,18 +56,18 @@
...
+            android:paddingRight="@dimen/layer_ui_margin_large">
 
             <LinearLayout
                 android:id="@+id/time_group"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
                 android:gravity="center"
                 android:orientation="horizontal"
-                android:paddingBottom="@dimen/layer_ui_message_item_time_group_padding_bottom"
+                android:paddingBottom="@dimen/layer_ui_message_item_time_group_padding"
```
thecombatwombat: This should have been either in the last PR or in a new PR

layer-atlas/src/main/java/com/layer/ui/message/MessageItemHeaderViewModel.java
andela-wakinsanmi: Thanks @thecombatwombat The `MessageItemHeaderViewModel` doesn't have an "Empty view", it only sets the empty string on the View if the View passed uses the ViewModel.

thecombatwombat: Let me refine that statement. The header knows nothing of anything being "empty". The word empty should not exist in the vocabulary of the header

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
andela-wakinsanmi: This is from a rebase of Layer_ui branch.

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
andela-wakinsanmi: Because this should be called only by MessageItemsList which is in the same package, since this is set only when the MessageItemsList wants to show the HeaderView

layer-atlas/src/main/java/com/layer/ui/message/EmptyMessageListView.java
```diff
@@ -0,0 +1,43 @@
...
+import android.databinding.DataBindingUtil;
+import android.support.annotation.Nullable;
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.widget.FrameLayout;
+
+import com.layer.ui.R;
+import com.layer.ui.databinding.UiEmptyMessageItemBinding;
+
+public final class EmptyMessageListView extends FrameLayout {
```
thecombatwombat: This class is not a list view, therefore should not have `listview` in its name. This class does not pertain to empty _messages_ either. 

Suggested name `EmptyMessageListHeaderView`

layer-atlas/src/main/java/com/layer/ui/message/EmptyMessageListView.java
```diff
@@ -0,0 +1,43 @@
...
+        super(context, attrs);
+        init();
+    }
+
+    public EmptyMessageListView(Context context, @Nullable AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+        init();
+    }
+
+    private void init() {
```
thecombatwombat: remove this `init` method. put the contents of this in 
`public EmptyMessageListView(Context context, @Nullable AttributeSet attrs, int defStyleAttr)` and have the other constructors reference this one.

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -79,18 +108,39 @@ public void onRefresh() {
         });
     }
 
+    private void removeEmptyHeaderView() {
+        int count = (mAdapter.getHeaderView() == null ? 0 : 1) + (mAdapter.getFooterView() == null ? 0 : 1);
+        if (mAdapter.getHeaderView() == mHeaderView && count != mAdapter.getItemCount()) {
+            mAdapter.setHeaderView(null);
+            mAdapter.setShouldShowHeader(false);
+            mAdapter.notifyDataSetChanged();
```
thecombatwombat: Please avoid calls to `notifyDataSetChanged` as much as possible. The only view you're removing here is the header view. The correct call is `notifyItemRemoved(index)`

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListView.java
```diff
@@ -79,18 +108,39 @@ public void onRefresh() {
         });
     }
 
+    private void removeEmptyHeaderView() {
+        int count = (mAdapter.getHeaderView() == null ? 0 : 1) + (mAdapter.getFooterView() == null ? 0 : 1);
+        if (mAdapter.getHeaderView() == mHeaderView && count != mAdapter.getItemCount()) {
```
thecombatwombat: Please explain the logic here

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
thecombatwombat: https://stackoverflow.com/questions/215497/in-java-difference-between-default-public-protected-and-private

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
```diff
@@ -46,6 +47,7 @@ public void setOnItemSwipeListener(SwipeableItem.OnItemSwipeListener<Message> on
         notifyChange();
     }
 
+    @Bindable
```
thecombatwombat: Why does this need to be bindable?

andela-wakinsanmi: @thecombatwombat I have it there because it's getter method

layer-atlas/src/main/java/com/layer/ui/message/EmptyMessageListHeaderView.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public EmptyMessageListHeaderView(Context context, @Nullable AttributeSet attrs) {
+        this(context, attrs, 0);
+    }
+
+    public EmptyMessageListHeaderView(Context context, @Nullable AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+
+        LayoutInflater inflater = (LayoutInflater) getContext().getSystemService(
```
thecombatwombat: This can be simplified to `LayoutInflater inflater = LayoutInflater.from(getContext());`

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
```diff
@@ -20,18 +19,17 @@
...
     public MessageItemsListViewModel(Context context, LayerClient layerClient,
-                                     ImageCacheWrapper imageCacheWrapper, DateFormatter dateFormatter,
+                                     ImageCacheWrapper imageCacheWrapper,
+                                     DateFormatter dateFormatter,
                                      IdentityFormatter identityFormatter) {
         mMessageItemsAdapter = new MessageItemsAdapter(context, layerClient,
                 imageCacheWrapper, dateFormatter, identityFormatter);
     }
 
-    @Bindable
```
thecombatwombat: Why have you removed these `@Bindable` methods? They are required

layer-atlas/src/main/java/com/layer/ui/message/MessageItemsListViewModel.java
```diff
@@ -20,18 +19,17 @@
...
     public MessageItemsAdapter getAdapter() {
         return mMessageItemsAdapter;
     }
 
-    @Bindable
```
thecombatwombat: Why have you removed these @Bindable methods? They are required

PR #281, opened 2017-10-26
Text Primitive + Message View + Message Container + MessageViewer

thecombatwombat:
This PR implements a few key rich messaging components;

1. `MessageModel` is a view model that can parse and interpret messages of a certain mime type, and supplies the class of the `MessageView` type that can render it.
2. `MessageView` can render messages parsed by a required `MessageModel`. It also indicates its preferred `MessageContainer` which may be ignored by the parent view holding this `MessageView`
3. `MessageContainer` is a view responsible for displaying a `MessageView` and optionally, any meta-data associated with the given message
4. `MessageViewer` is the outermost view that contains all rendering related to rich messages. Is responsible for determining which `MessageModel` to use when rendering a given message, and which `MessageContainer` should be used to render the generated `MessageView`. `MessageViewer` is a child view in a message item. 
5.`MessageModelManager` allows for the registry of `MessageModel` classes against the root mime type of the message they can parse. 
6. `BinderRegistry` determines with the help of `MessagePartUtils`, whether a message is to be displayed using the new rich messaging mechanisms, or to fall back to the old `CellFactories`

`MessageModelManager` and `MessageViewer` impose constraints on the constructors that can be used on `MessageModel` and `MessageView` instances respectively. This is done to keep a simple class structure and avoid the complexities of exposing factories or other object generation code. As this evolves, we might re-visit those constraints and/or associated mechanisms. 

In addition to the above basic components, the PR also includes an implementation for `TextMessageView`, which can display text messages along with optional meta-data.


layer-atlas/src/main/java/com/layer/ui/message/binder/BinderRegistry.java
```diff
@@ -163,4 +168,28 @@ public void setCellFactories(List<CellFactory> cellFactories) {
     public MessageCell getMessageCellForViewType(int viewType) {
         return mCellTypesByViewType.get(viewType);
     }
+
+    //==============================================================================================
+    //  XDK Message Binding related methods
+    //==============================================================================================
+
+    private void initMessageTypeModelRegistry() {
+        mMessageModelManager.registerModel("application/vnd.layer.text+json", TextMessageModel.class);
```
andela-wakinsanmi: Can't we make the string "application/vnd.layer.text+json" a constant, giving it a descriptive name

layer-atlas/src/main/java/com/layer/ui/message/container/StandardMessageContainer.java
```diff
@@ -0,0 +1,70 @@
...
+        super(context, attrs, defStyleAttr);
+        LayoutInflater inflater = LayoutInflater.from(getContext());
+        mBinding = UiStandardMessageContainerBinding.inflate(inflater, this, true);
+        mBinding.uiStandardMessageContainerTitle.setVisibility(GONE);
+        mBinding.uiStandardMessageContainerSubtitle.setVisibility(GONE);
+    }
+
+    public void setMessageView(MessageView view) {
+        mMessageView = view;
+        view.setLayoutParams(new ViewGroup.LayoutParams(
```
andela-wakinsanmi: Is there a particular reason why we are overriding the layout params of the View passed here?

layer-atlas/src/main/java/com/layer/ui/message/container/StandardMessageContainer.java
```diff
@@ -0,0 +1,70 @@
...
+                ViewGroup.LayoutParams.WRAP_CONTENT,
+                ViewGroup.LayoutParams.MATCH_PARENT));
+        mBinding.uiStandardMessageContainerContentView.addView(view);
+    }
+
+    @Override
+    public <T extends MessageModel> void setMessageModel(T model) {
+        mMessageView.setMessageModel(model);
```
andela-wakinsanmi: Is it possible at any point in time for setMessageModel() to be called before setMessageView? We may want to have a null check here

layer-atlas/src/main/java/com/layer/ui/message/binder/BinderRegistry.java
thecombatwombat: Good point. Still trying to come.up with a better way to do this, but short of that,a constant would be helpful

layer-atlas/src/main/java/com/layer/ui/message/container/StandardMessageContainer.java
thecombatwombat: The container manages the layout of the message content, hence is responsible for setting it.

layer-atlas/src/main/java/com/layer/ui/message/container/StandardMessageContainer.java
thecombatwombat: If someone calls `setMessageModel()` before setting a view, then it _should_ fail.  The resulting exception would warn the developer that they've done something wrong. A null check that eats the error would leave them with no indication as to what is going on.

layer-atlas/src/main/java/com/layer/ui/message/container/MessageContainer.java
```diff
@@ -0,0 +1,31 @@
...
+import android.support.annotation.AttrRes;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.util.AttributeSet;
+import android.widget.FrameLayout;
+
+import com.layer.ui.message.model.MessageModel;
+import com.layer.ui.message.view.MessageView;
+
+public abstract class MessageContainer extends FrameLayout {
```
smpete: Should this be an interface instead?

layer-atlas/src/main/java/com/layer/ui/message/messagetypes/text/TextSender.java
```diff
@@ -2,6 +2,7 @@
 
 import android.content.Context;
 
+import com.google.gson.JsonObject;
```
smpete: Not needed

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
```diff
@@ -0,0 +1,136 @@
...
+    @Override
+    public void onProgressComplete(MessagePart messagePart, Operation operation) {
+        decrementDownloadingPartCounter();
+        parse(messagePart);
+        notifyChange();
+    }
+
+    @Override
+    public void onProgressError(MessagePart messagePart, Operation operation, Throwable throwable) {
+        decrementDownloadingPartCounter();
```
smpete: Should this notify changes as well? I'd think it would need to so the `isDownloadingParts` value remains correct.

Similarly, how are errors propagated down to the model implementations?

layer-atlas/src/main/java/com/layer/ui/message/model/TextMessageModel.java
```diff
@@ -0,0 +1,83 @@
...
+        String data = new String(messagePart.getData());
+        JsonObject jsonObject = mJsonParser.parse(data).getAsJsonObject();
+        mText = jsonObject.get("text").getAsString().trim();
+        mSubtitle = jsonObject.has("description") ? jsonObject.get("description").getAsString() : null;
+        mTitle = jsonObject.has("title") ? jsonObject.get("title").getAsString() : null;
+    }
+
+    @Bindable
+    public String getText() {
+        return !TextUtils.isEmpty(mText) ? mText : null;
```
smpete: Why do we need these isEmpty checks? Do we want to make sure we never return an empty String?

layer-atlas/src/main/java/com/layer/ui/message/view/MessageView.java
```diff
@@ -0,0 +1,27 @@
...
+
+import android.content.Context;
+import android.support.annotation.Nullable;
+import android.util.AttributeSet;
+import android.widget.FrameLayout;
+
+import com.layer.ui.message.container.MessageContainer;
+import com.layer.ui.message.model.MessageModel;
+
+public abstract class MessageView<VIEW_MODEL extends MessageModel> extends FrameLayout {
```
smpete: Similar question as before, should this be an interface instead?

thecombatwombat: Need to restrict constructors, so no, that can't be done with just an interface

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -366,12 +367,12 @@ public int getItemViewType(int position) {
             return createHeaderViewHolder(parent);
         } else if (viewType == mBinderRegistry.VIEW_TYPE_FOOTER) {
             return createFooterViewHolder(parent);
-        } else if (viewType == mBinderRegistry.VIEW_TYPE_CARD) {
-            return createCardMessageItemViewHolder(parent);
-        } else { // Is a legacy view type
+        } else if (viewType >= mBinderRegistry.VIEW_TYPE_LEGACY_START && viewType <= mBinderRegistry.VIEW_TYPE_LEGACY_END) {
```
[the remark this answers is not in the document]
thecombatwombat: Yeah, this class needs some cleaning up... might be worth turning them into `isLegacyViewType(int viewType)` style methods that move the range check into the `BinderRegistry`, and remove the ability to set legacy start and end

layer-atlas/src/main/java/com/layer/ui/message/container/MessageContainer.java
thecombatwombat: Need to be able to add subviews, so no.

layer-atlas/src/main/java/com/layer/ui/message/model/TextMessageModel.java
thecombatwombat: Need to be able to hide or show stuff depending on whether these strings are empty or not

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
```diff
@@ -0,0 +1,136 @@
...
+    private IdentityFormatter mIdentityFormatter;
+    private DateFormatter mDateFormatter;
+
+    private final Context mContext;
+    private final LayerClient mLayerClient;
+
+    private Message mMessage;
+
+    public MessageModel(Context context, LayerClient layerClient) {
+        mIdentityFormatter = new IdentityFormatterImpl(context);
```
andela-wakinsanmi: Should we allow the user to quickly swap these default objects with their own implementation?

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
thecombatwombat: Model implementations are free to override and call `super` and if required, notify changes. Lets see a use case for it come up and we do more complex message types, and we'll revisit accordingly.

sriamar:
please add a description?

layer-atlas/src/main/java/com/layer/ui/message/MessagePartUtils.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.layer.sdk.messaging.Message;
+import com.layer.sdk.messaging.MessagePart;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class MessagePartUtils {
```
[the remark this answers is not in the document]
sriamar: We should be a bit more careful about moving this to core. Let's chat about this when we get there.

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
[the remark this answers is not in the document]
thecombatwombat: This is best dealt with in another PR, will fix as a standalone task.

PR #286, opened 2017-11-14
Task/and 1011 image primitive

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/message/container/MessageContainer.java
```diff
@@ -21,11 +27,56 @@ public MessageContainer(@NonNull Context context, @Nullable AttributeSet attrs)
...
+
+    protected void setCornerClippingPath(Path cornerClippingPath) {
+        mCornerClippingPath = cornerClippingPath;
+    }
+
     protected abstract <T extends MessageModel> void setContentBackground(T model);
+
+    @Override
+    protected void dispatchDraw(Canvas canvas) {
+        if (mCornerRadius > 0) {
```
smpete: So suppress this draw if the corner radius is 0?

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+import android.databinding.BaseObservable;
+import android.support.annotation.Dimension;
+import android.support.annotation.Nullable;
+
+import com.layer.ui.util.display.DisplayUtils;
+
+import java.util.Map;
+
+public class ImageMessageMetadata extends BaseObservable {
+    public String mTitle;
```
smpete: Should be private?

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageModel.java
```diff
@@ -0,0 +1,174 @@
...
+        return false;
+    }
+
+    /*
+    * Private methods
+    */
+
+    private ImageMessageMetadata parseRootMessagePart(MessagePart messagePart) {
+        JsonReader reader = new JsonReader(new InputStreamReader(messagePart.getDataStream()));
+        return mGson.fromJson(reader, ImageMessageMetadata.class);
```
smpete: How does this end up notifying the `ImageMessageMetadata` that its properties have changed?

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageModel.java
```diff
@@ -0,0 +1,174 @@
...
+        return mMetadata;
+    }
+
+    /*
+    * Setters, getters, bindings
+    */
+
+    public ImageCacheWrapper getImageCacheWrapper() {
+        if (sImageCacheWrapper == null) {
+            sImageCacheWrapper = new PicassoImageCacheWrapper(new Picasso.Builder(getContext())
```
smpete: Why should this class be responsible for created an image cache wrapper? Shouldn't there be an error if none is set?

layer-atlas/src/main/java/com/layer/ui/util/display/DisplayUtils.java
```diff
@@ -0,0 +1,16 @@
+package com.layer.ui.util.display;
+
+import android.content.res.Resources;
+import android.util.DisplayMetrics;
+
+public class DisplayUtils {
```
smpete: Do we really need this as a utils class? I've usually just used TypedValue.applyDimension() for dp -> px conversion. I don't see where pxToDp is used...

layer-atlas/src/main/java/com/layer/ui/util/imagecache/PicassoImageCacheWrapper.java
```diff
@@ -100,55 +100,88 @@ public void resumeTag(String picassoTag) {
...
 
-        if (imageRequestParameters.getRotateAngleTo() != 0) {
-            requestCreator.rotate(imageRequestParameters.getRotateAngleTo());
+        if (imageRequestParameters.getTargetWidth() > 0 && imageRequestParameters.getTargetHeight() > 0) {
+            requestCreator.resize(imageRequestParameters.getTargetWidth(), imageRequestParameters.getTargetHeight());
         }
 
-        if (imageRequestParameters.shouldScaleDownTo()) {
+        float rotationAngle = imageRequestParameters.getRotationDegrees() +
+                imageRequestParameters.getExifRotationInDegrees(imageView.getContext());
```
smpete: Isn't the `loadImage` method safe to call from any thread? This can cause a disk read when it queries the `ContentResolver`

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageRequestParameters.java
```diff
@@ -1,121 +1,418 @@
...
-            mResizeHeightTo = resizeHeightTo;
+
+            // Clear any resource id or url that may be set
+            mResourceId = 0;
+            mUrl = null;
+            return this;
+        }
+
+        public Builder url(@NonNull String url) {
+            if (url == null || url.isEmpty()) {
```
smpete: Maybe validation should happen during the `build()` call instead. If the the `url` method isn't called then this validation isn't performed

layer-atlas/src/main/java/com/layer/ui/message/container/MessageContainer.java
thecombatwombat: 🤦‍♂️ I'll add the else block

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageModel.java
thecombatwombat: This class should not be responsible for creating the `ImageCacheWrapper`, but neither should its caller (some container class) be responsible for figuring out every crazy dependency that is required by any `MessageView` type. 

I should have left a comment, but this is in there as a stopgap, until we have the DI/Dagger/Whatever discussion in all its glory.

layer-atlas/src/main/java/com/layer/ui/util/display/DisplayUtils.java
thecombatwombat: 🆒  wrt using `TypedValue.applyDimension` for dp -> px 
Although I'm not wild about having the metadata object deal with `DisplayMetrics` and stuff. I wonder if that conversion is best done by whoever is using this value, rather than in the metadata class. Thoughts?

I'll need px to dp for when image messages are created from an android phone and I need to create a metadata object for the root message part.

layer-atlas/src/main/java/com/layer/ui/util/imagecache/ImageRequestParameters.java
thecombatwombat: Well if the `url` method is not called, then the value of `mUrl` is null, and that is perfectly fine, as long as either `mUri` or `mResourceId` have valid values.  So the only possible place for that validation is `build`

layer-atlas/src/main/java/com/layer/ui/util/imagecache/PicassoImageCacheWrapper.java
thecombatwombat: `exif` orientation is a bit of a thorny issue right now between how server and mobile clients are handling it. Lukasz and I are not wild about exif coming from 2 different places, while still trying to handle exif rotation in some consistent manner. 

We'd like to always respect the `exif` rotation in the `jpeg` (but have no guarantees of the file being or having `exif` data in it) 

I don't have a straight answer for how Picasso handles exif orientation. I know it does not read exif from the jpeg if it is coming off the network, but I'm not sure if it correctly handles the exif from files and content uris. 

Hence the checks. 

As for the question of disk reads off the UI thread, I dunno... ideas?

smpete: Well we'd need to offload that into a thread somehow. It'll likely require a fair bit of work. I'm okay filing a bug for this and moving on for now.

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageComposer.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private Context mContext;
+    private LayerClient mLayerClient;
+
+    public ImageMessageComposer(Context context, LayerClient layerClient) {
+        mContext = context;
+        mLayerClient = layerClient;
+    }
+
+    public abstract Message newImageMessage(@NonNull Uri imageUri) throws IOException;
```
smpete: These `newImageMessage` methods are making disk reads. We need to fix the threading here.

layer-atlas/src/main/java/com/layer/ui/message/action/ActionHandler.java
```diff
@@ -0,0 +1,28 @@
+package com.layer.ui.message.action;
+
+import android.content.Context;
+import android.support.annotation.NonNull;
+
+import com.google.gson.JsonObject;
+import com.layer.sdk.LayerClient;
+
+public abstract class ActionHandler {
```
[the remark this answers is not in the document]
thecombatwombat: Now that you mention it, I concur. Lemme see if I can come up with a new name/if it mucks with what we have in the spec

layer-atlas/src/main/java/com/layer/ui/message/messagetypes/threepartimage/GallerySender.java
```diff
@@ -96,7 +113,7 @@ public boolean onActivityResult(Activity activity, int requestCode, int resultCo
             Identity me = getLayerClient().getAuthenticatedUser();
             String myName = me == null ? "" : Util.getDisplayName(me);
             Uri uri = data.getData();
-            Message message = ThreePartImageUtils.newThreePartImageMessage(activity, getLayerClient(), uri);
+            Message message = mImageMessageComposer.newImageMessage(uri);
```
smpete: URI might be null here

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageComposer.java
```diff
@@ -0,0 +1,162 @@
...
+            return previewBitmap;
+        } else {
+            return sampledBitmap;
+        }
+    }
+
+    protected long getFileSizeFromUri(Context context, Uri uri) {
+        Cursor cursor = context.getContentResolver().query(uri,
+                null, null, null, null);
+        cursor.moveToFirst();
```
smpete: Cursor can be null here

layer-atlas/src/main/java/com/layer/ui/message/image/ImageMessageComposer.java
```diff
@@ -0,0 +1,162 @@
...
+
+    protected void writeStreamToFile(String filePath, InputStream inputStream) throws IOException {
+        OutputStream stream = new BufferedOutputStream(new FileOutputStream(filePath));
+        int bufferSize = 1024;
+        byte[] buffer = new byte[bufferSize];
+        int len = 0;
+        while ((len = inputStream.read(buffer)) != -1) {
+            stream.write(buffer, 0, len);
+        }
+        if (stream != null)
```
smpete: The `stream` is never null

layer-atlas/src/main/java/com/layer/ui/message/action/ActionHandler.java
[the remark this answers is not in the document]
thecombatwombat: Leaving it be for now, can change if we come up with something better

PR #287, opened 2017-11-15
GIF Integration using Tenor

ctrl-alt-del:
This is a GIF integration using Tenor and some classes that I created.


layer-atlas/build.gradle
```diff
@@ -70,6 +74,16 @@ dependencies {
     androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
+
+    compile(name: 'tenor-android-sdk', ext: 'aar') {
```
smpete: This does not work when compiling Atlas-Android as a project. Tested with Atlas-Android-Messenger and it gives the error: 

> Failed to resolve: :tenor-android-sdk:

 Can this AAR be included without requiring additions to the main applications build file?

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
```diff
@@ -45,26 +47,49 @@
...
+import com.layer.atlas.tenor.adapters.OnDismissPopupWindowListener;
+import com.layer.atlas.tenor.presenters.impl.GifPresenter;
+import com.layer.atlas.tenor.rvitem.ResultRVItem;
+import com.layer.atlas.tenor.threepartgif.GifSender;
+import com.layer.atlas.tenor.views.IKeyboardView;
 import com.layer.atlas.util.EditTextUtil;
 import com.layer.sdk.LayerClient;
 import com.layer.sdk.listeners.LayerTypingIndicatorListener;
 import com.layer.sdk.messaging.Conversation;
+import com.tenor.android.sdk.constants.StringConstant;
```
smpete: What source was this library produced from? The Tenor GitHub project has everything under the package `com.tenor.android.sdk.core`.

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
```diff
@@ -86,6 +87,7 @@ public void bindCellHolder(CellHolder cellHolder, final TextInfo parsed, Message
                 cellHolder.mProgressBar.show();
             }
         }
+        SmartGifsUtils.update(textMessage, specs.position);
```
thecombatwombat: Please write a separate class with the changes you wish to have here.

Our developers depend on the current behavior of this class, and may not wish to use Tenor. We cannot force this functionality on them.

layer-atlas/src/main/java/com/layer/atlas/tenor/adapters/GifAdapter.java
```diff
@@ -0,0 +1,124 @@
...
+import com.tenor.android.sdk.models.Result;
+import com.tenor.android.sdk.rvwidgets.AbstractRVItem;
+import com.tenor.android.sdk.rvwidgets.ListRVAdapter;
+import com.tenor.android.sdk.rvwidgets.StaggeredGridLayoutItemViewHolder;
+import com.tenor.android.sdk.utils.AbstractListUtils;
+
+import java.util.List;
+import java.util.Map;
+
+public class GifAdapter<CTX> extends ListRVAdapter<CTX, AbstractRVItem, StaggeredGridLayoutItemViewHolder<CTX>> {
```
thecombatwombat: Why is the `<CTX>` generic required here? If a `Context` is expected for this, then this generic can be removed and the class can be defined as 

`+public class GifAdapter extends ListRVAdapter<Context, AbstractRVItem, StaggeredGridLayoutItemViewHolder<Context>>`

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
ctrl-alt-del: The source of this library is https://github.com/Tenor-Inc/tenor-android-core, this specific version used in the PR is a earlier version of source, and thus, you see `com.tenor.android.sdk` instead of `com.tenor.android.core`;

ctrl-alt-del:
hey @thecombatwombat , good to hear back from you.

I read your questions and understand you concerns, and I appreciate that you are going to keep our changes in a separate branch called `integrations/tenor` for now.  I will address your comments accordingly.

For the questions that you listed in above, here are my thoughts:

Correct me if I am wrong, I read your `MessageSender`, `Conversation` and `Message` classes again and noticed that you abstracted the network requesting/receiving action using interface.  I maybe able to take the similar approach to exclude the `Retrofit`, or is there any libraries that you prefer, since the built-in `Apache HTTP` is removed on API 23, (https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-apache-http-client).

For `Glide`, it is more complicated.  The `Piccaso` library you are using doesn't support loading GIFs and they are not planning to in the foreseeable future (https://github.com/square/picasso/issues/700).  Therefore, we need to use something else, and `Glide` is the widely recommended library.  Of course, I can abstract it as a interface as well, so that the dependency requirement won't be living in `Atlas-Android`, but rather living in the `Atlas-Android-Messenger`.  

Would abstracting those libraries out of `Atlas-Android` makes it easier for you to merge into `master`?

For the dex count, I think you can use ProGuard to exclude all the dependency that you don't need in you specific use case.  We use most of the library on our `com.tenor.android.core`, so we don't ProGuard them.

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
```diff
@@ -45,26 +47,49 @@
...
     private LayerClient mLayerClient;
     private Conversation mConversation;
 
     private TextSender mTextSender;
+    private GifSender mGifSender;
     private ArrayList<AttachmentSender> mAttachmentSenders = new ArrayList<AttachmentSender>();
     private MessageSender.Callback mMessageSenderCallback;
 
     private PopupWindow mAttachmentMenu;
+    private RecyclerView mGifsRV;
```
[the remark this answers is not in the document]
ctrl-alt-del: ok will do.

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
```diff
@@ -45,26 +47,49 @@
...
+import java.util.List;
 
-public class AtlasMessageComposer extends FrameLayout {
+import retrofit2.Call;
+
+public class AtlasMessageComposer extends FrameLayout implements IKeyboardView {
     private EditText mMessageEditText;
     private Button mSendButton;
     private ImageView mAttachButton;
+    private ImageView mOpenGifsRVButton;
```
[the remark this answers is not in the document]
ctrl-alt-del: sure will do.

layer-atlas/src/main/java/com/layer/atlas/AtlasMessageComposer.java
```diff
@@ -91,16 +122,97 @@ public AtlasMessageComposer(Context context, AttributeSet attrs, int defStyle) {
...
+    }
+
     /**
      * Prepares this AtlasMessageComposer for use.
      *
      * @return this AtlasMessageComposer.
      */
     public AtlasMessageComposer init(LayerClient layerClient) {
-        LayoutInflater.from(getContext()).inflate(R.layout.atlas_message_composer, this);
+        LayoutInflater.from(getContext()).inflate(R.layout.tenor_message_composer, this);
```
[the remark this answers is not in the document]
ctrl-alt-del: Ok will do

layer-atlas/src/main/java/com/layer/atlas/messagetypes/text/TextCellFactory.java
ctrl-alt-del: Ok, will create a `TenorTextCellFactory` and subclass the `TextCellFactory`

layer-atlas/src/main/java/com/layer/atlas/tenor/adapters/GifAdapter.java
ctrl-alt-del: I was expecting `GifAdapter<TenorMessageComposer>`, will fix it.

thecombatwombat:
@ctrl-alt-del Hey... Indeed Apache HTTP was removed in API 23, however, I do believe the `HttpUrlConnection` class is indeed available all the way back to API 1. 

As far as Glide is concerned, we know that Picasso does not support animated gifs, so Glide might be the best option here. 

I do understand that using only standard android APIs for networking/image loading is not always easy. 

So for now, as long as the Tenor integration is in another branch, we're okay with you bringing in an additional dependency or two. We'll have to take another long and hard look at the same problems when we try and integrate Tenor into the mainline, but hopefully by then we would have done some work to make integrations like this easier and more modular. 

Please let me know if you have more questions!

Cheers,
Archit

ctrl-alt-del:
Greeting! @thecombatwombat,  thanks for your replies, they are helpful.

Per our conversation earlier, I have moved most of the implementations out of the `Atlas-Android`; it doesn't require additional dependencies on `Atlas-Android` now.  Instead, the dependencies are now located in the `Atlas-Android-Messenger`.

You can check out the compare diff of `Atlas-Android` in [here](https://github.com/layerhq/Atlas-Android/compare/master...ctrl-alt-del:feature/dev_gif-integration-with-abstraction?diff=split&expand=1&name=feature%2Fdev_gif-integration-with-abstraction).

You can check out the compare diff of `Atlas-Android-Messenger` in [here](https://github.com/layerhq/Atlas-Android-Messenger/compare/master...ctrl-alt-del:feature/dev_tenor-gif-integration?expand=1).  I built and tested your demo messenger and it works.

Meanwhile, I found an issue when I was testing your `Atlas-Android-Messenger` with the identity provider I deployed to Heroku using your latest script.  It always return 404 on authenticating a user.  It turned out that it was caused by a mismatch JSON parameter structure and a wrong authenticate `url`; I patched it and you can check out the diff in [here](https://github.com/layerhq/Atlas-Android-Messenger/compare/master...ctrl-alt-del:feature/dev_fix-404-response-on-identity-provider-authentication?expand=1).

I can create PRs for my changes if you prefer.  Let me know your thoughts.

Best,
Jiyang

thecombatwombat:
@ctrl-alt-del Please open a PR against the `integration/tenor` branch in both `Atlas` and `Atlas-Android-Messenger` 

The diffs mostly look okay, but we'll take another look before we hit approve...

ctrl-alt-del:
@thecombatwombat thank you.  Here they are: [Atlas-Android](https://github.com/layerhq/Atlas-Android/pull/298) and [Atlas-Android-Messenger](https://github.com/layerhq/Atlas-Android-Messenger/pull/111).

smpete:
Closing this as #298 now represents this change

PR #293, opened 2017-11-28
Fix ConversationsRecyclerView is not scrolled

andela-wakinsanmi:


layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
```diff
@@ -39,8 +39,23 @@
...
-
     private ConversationStyle conversationStyle;
+    private LinearLayoutManager mLinearLayoutManager;
+
+    private RecyclerView.AdapterDataObserver mDataObserver = new RecyclerView.AdapterDataObserver() {
+
+        @Override
+        public void onItemRangeMoved(int fromPosition, int toPosition, int itemCount) {
+            super.onItemRangeMoved(fromPosition, toPosition, itemCount);
+            scrollTo(toPosition);
```
smpete: If an item moved from position 20 to position 19 would we want to scroll to position 20?

layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
```diff
@@ -39,8 +39,23 @@
...
+        }
+
+        @Override
+        public void onItemRangeInserted(int positionStart, int itemCount) {
+            super.onItemRangeInserted(positionStart, itemCount);
+            scrollTo(positionStart);
```
smpete: Similarly, if an item was inserted at position 10 would we want to scroll to position 10?

layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
```diff
@@ -82,15 +97,24 @@ public void setAdapter(Adapter adapter) {
 
     public AtlasConversationsRecyclerView addCellFactories (AtlasCellFactory... cellFactories) {
         mAdapter.addCellFactories(cellFactories);
+        mAdapter.registerAdapterDataObserver(mDataObserver);
```
smpete: Why is this done here and not in the `init()`?

layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
andela-wakinsanmi: No, I wanted to make that scroll to position(0), again if an item is moved at position 20 we don't want to scroll to position(0). Hence, the scrollTo method checks the Visible first item and ensure it's less than 3 before the scrolling happens. If an item is moved at position 20, the scroll should not happen as far as the firstVisibleItem is greater than 2. Any thought on a better way to fix this?

smpete: Right. We'd likely want to scroll here only if the item moved to position 0. Should be a simple check

layer-atlas/src/main/java/com/layer/atlas/AtlasConversationsRecyclerView.java
```diff
@@ -85,12 +101,20 @@ public AtlasConversationsRecyclerView addCellFactories (AtlasCellFactory... cell
         return this;
     }
 
+    private void scrollTo(int toPosition) {
```
smpete: This should be scoped inside the `AdapterDataObserver` since that's the only object to use this method. Also a more descriptive name couldn't hurt. Or extract the check to a method that returns a boolean

PR #300, opened 2017-12-12
Task/and 1012 location primitive

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+            return false;
+        }
+        if (Log.isLoggable(Log.VERBOSE)) Log.v("Getting fresh location");
+        LocationRequest r = new LocationRequest()
+                .setNumUpdates(1)
+                .setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)
+                .setExpirationDuration(10000)
+                .setMaxWaitTime(10000);
+        try {
+            LocationServices.FusedLocationApi.requestLocationUpdates(sGoogleApiClient, r, listener);
```
smpete: What happens if the client is disconnected here?

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    @RequiresPermission(PERMISSION)
+    private static boolean getFreshLocation(LocationListener listener) {
+        if (sGoogleApiClient == null) {
+            if (Log.isLoggable(Log.ERROR)) Log.e("GoogleApiClient not initialized");
+            return false;
+        }
+        if (Log.isLoggable(Log.VERBOSE)) Log.v("Getting fresh location");
+        LocationRequest r = new LocationRequest()
```
smpete: Use constants for duration and wait time numbers. I'm indifferent on the number of updates if it's a constant or inlined.

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
smpete: Also, this is method of getting a location is deprecated. See https://developers.google.com/android/reference/com/google/android/gms/location/LocationServices#FusedLocationApi

We should probably take this opportunity to update play services to the latest

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+        return false;
+    }
+
+    @RequiresPermission(PERMISSION)
+    @Override
+    public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+        if (requestCode != PERMISSION_REQUEST_CODE) return;
+        if (grantResults.length == 0 || grantResults[0] != PackageManager.PERMISSION_GRANTED) {
+            if (Log.isLoggable(Log.VERBOSE)) Log.v("Location permission denied");
```
smpete: Do we need to eventually convey this to the developer/user?

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Asynchronously requests a fresh location and sends a location Message.
+     */
+    @RequiresPermission(PERMISSION)
+    @Override
+    public boolean requestSend() {
+        Activity activity = mActivity.get();
+        if (activity == null) return false;
+        if (Log.isLoggable(Log.VERBOSE)) Log.v("Sending location");
+        if (checkSelfPermission(activity, PERMISSION) != PackageManager.PERMISSION_GRANTED) {
```
smpete: I'd prefer to see this method prefixed with `ContextCompat.`. This helps readers know that this method is external to the class

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+        if (requestCode != PERMISSION_REQUEST_CODE) return;
+        if (grantResults.length == 0 || grantResults[0] != PackageManager.PERMISSION_GRANTED) {
+            if (Log.isLoggable(Log.VERBOSE)) Log.v("Location permission denied");
+            return;
+        }
+        getFreshLocation(new SenderLocationListener(this));
+    }
+
+    /**
+     * Asynchronously requests a fresh location and sends a location Message.
```
smpete: Javadoc is missing @return

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+            requestPermissions(activity, PERMISSION_REQUEST_CODE, PERMISSION);
+            return true;
+        }
+        return getFreshLocation(new SenderLocationListener(this));
+    }
+
+    @Override
+    public boolean onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
+        if (requestCode != ACTIVITY_REQUEST_CODE) return false;
+        init(activity);
```
smpete: Why does this re-init?

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+        return true;
+    }
+
+    private static class SenderLocationListener implements LocationListener {
+        private final WeakReference<CurrentLocationSender> mCurrentLocationSenderWeakReference;
+
+        public SenderLocationListener(CurrentLocationSender locationsender) {
```
smpete: Camel case the parameter

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+            mCurrentLocationSenderWeakReference = new WeakReference<CurrentLocationSender>(locationsender);
```
smpete: Use diamond operator

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
smpete: Visibility should be private

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageMetadata.java
```diff
@@ -0,0 +1,236 @@
...
+    public JSONObject getCustomData() {
+        return mCustomData;
+    }
+
+    @Nullable
+    public Action getAction() {
+        return mAction;
+    }
+
+    // TODO : Understand how to Localize display of addresses
```
smpete: File a Jira ticket for this task and put the ticket number in the comment for tracking

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
```diff
@@ -0,0 +1,172 @@
...
+import com.layer.ui.util.imagecache.ImageRequestParameters;
+import com.layer.ui.util.imagecache.PicassoImageCacheWrapper;
+import com.layer.ui.util.imagecache.requesthandlers.MessagePartRequestHandler;
+import com.layer.ui.util.json.AndroidFieldNamingStrategy;
+import com.squareup.picasso.Picasso;
+
+import java.io.InputStreamReader;
+
+public class LocationMessageModel extends MessageModel {
+    public static final String ACTION_EVENT_OPEN_MAP = "open-map";
```
smpete: Do these need to be public constants?

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
```diff
@@ -0,0 +1,172 @@
...
+
+    @Override
+    public Class<LocationMessageView> getRendererType() {
+        return LocationMessageView.class;
+    }
+
+    @Override
+    protected void parse(MessagePart messagePart) {
+        JsonReader reader = new JsonReader(new InputStreamReader(messagePart.getDataStream()));
+        mMetadata = mGson.fromJson(reader, LocationMessageMetadata.class);
```
smpete: We should do something with the json exceptions this can throw

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
```diff
@@ -0,0 +1,172 @@
...
+        return R.color.ui_location_message_background;
+    }
+
+    @Override
+    public boolean getHasContent() {
+        return mMetadata != null;
+    }
+
+    public ImageCacheWrapper getImageCacheWrapper() {
+        if (sImageCacheWrapper == null) {
```
smpete: What was decided on creating an image cache wrapper here?

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        return sImageCacheWrapper;
+    }
+
+    public static void setImageCacheWrapper(ImageCacheWrapper imageCacheWrapper) {
+        sImageCacheWrapper = imageCacheWrapper;
+    }
+
+    @Nullable
+    public ImageRequestParameters getMapImageRequestParamters() {
```
smpete: Typo

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
```diff
@@ -0,0 +1,172 @@
...
+            } else if (mMetadata.getFormattedAddress() != null) {
+                url.append(mMetadata.getFormattedAddress());
+            }
+
+            // Set dimensions
+            // Google Static Map API has max dimension 640
+            int mapWidth = (int) getContext().getResources().getDimension(R.dimen.layer_ui_location_message_map_width);
+            int mapHeight = (int) Math.round((double) mapWidth / GOLDEN_RATIO);
+
+            url.append("&size=").append(mapWidth).append("x").append(mapHeight);
```
smpete: If lat/long and formatted address are both null then the URL will be appended with `&markers=&size=...`. Is that okay?

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.gson.JsonObject;
+import com.layer.sdk.LayerClient;
+
+import java.util.Locale;
+
+public class GoogleMapsOpenMapActionHandler extends ActionHandler {
+
+    public static final String ACTION_EVENT = "open-map";
+
+    public static final String KEY_ADDRESS = "address";
```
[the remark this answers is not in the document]
thecombatwombat: I would like these to be discoverable, so basically if anyone wants to see what fields/keys they need to include to get this action handler to fire, its available to them through the IDE. No having to dig into docs. 

Not opposed to making them private/package visible, but not sure if that nets us much.

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
```diff
@@ -0,0 +1,78 @@
...
+            if (data.has(KEY_ADDRESS)) {
+                googleMapsUri = constructGoogleMapsUri(longitude, latitude,
+                        data.get(KEY_ADDRESS).getAsString(), markerTitle, zoom);
+            } else {
+                googleMapsUri = constructGoogleMapsUri(latitude, longitude, markerTitle, zoom);
+            }
+
+            Intent openMapsIntent = new Intent(Intent.ACTION_VIEW, googleMapsUri);
+
+            if (openMapsIntent.resolveActivity(context.getPackageManager()) != null) {
```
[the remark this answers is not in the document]
thecombatwombat: Actually this is a great case for a (functional) no-op. There is no activity available to handle said intent, so a no-op is the best thing you can do. 

We could toast an error message, I'm not opposed to that, but as I said, functionally, I think a no-op is the best option here.

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+import com.layer.ui.message.messagetypes.AttachmentSender;
+import com.layer.ui.util.Log;
+import com.layer.ui.util.json.AndroidFieldNamingStrategy;
+
+import java.lang.ref.WeakReference;
+
+import static android.support.v4.content.ContextCompat.checkSelfPermission;
+
+/**
+ * CurrentLocation creates a LocationMessage with the user's latest location.
```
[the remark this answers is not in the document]
thecombatwombat: I'll change that to Location Message since that is kinda proper noun-ish in our terminology. Lemme know if you think otherwise

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+ * Google's fused location API is used for gathering location at send time and may trigger a dialog
+ * for updating Google Play Services.  Requires `Manifest.permission.ACCESS_FINE_LOCATION`
+ * for getting device location.
+ */
+public class CurrentLocationSender extends AttachmentSender {
+    private static final String PERMISSION = Manifest.permission.ACCESS_FINE_LOCATION;
+    private static final int ACTIVITY_REQUEST_CODE = 30;
+    public static final int PERMISSION_REQUEST_CODE = 31;
```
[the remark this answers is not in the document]
thecombatwombat: 🤔 _I_ think this should be public since developers should be able to see/discover this code just in case it conflicts with a permission request code they might be using themselves. Now, they can't change this code (no api for that here), but they might be able to adjust their own codes to account for this. 

The `LocationSender` also does this, and if we change this, should we consider the same change on that class as well?

Thoughts?

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+                    .addConnectionCallbacks(googleApiCallbacks)
+                    .addOnConnectionFailedListener(googleApiCallbacks)
+                    .addApi(LocationServices.API)
+                    .build();
+            sGoogleApiClient.connect();
+            return;
+        }
+
+        // If the correct Google Play Services are not available, redirect to proper solution.
+        if (GooglePlayServicesUtil.isUserRecoverableError(errorCode)) {
```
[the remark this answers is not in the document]
thecombatwombat: Fixed. Moved over to `GoogleApiAvailability.getInstance().isUserResolvableError(errorCode)`

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
thecombatwombat: I wonder whats up with AS, not seeing that as deprecated here. I can take a look at updating Play Services.

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: The DI discussion was something we decided to punt on until we saw more use cases/how things fell.

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: Ack. We should just return null in that case and log things. 

More extreme would be throw an `IllegalArgumentException`. Thoughts?

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: I think this class should declare the action event it will respond to by default and have that be discoverable. 

Open to making this private. Thoughts?

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: Those are runtime exceptions, and should not be caught here. Those are caused by something being wrong in the json, and catching those exceptions would not help us recover from those errors in any meaningful way. 

Those exceptions should happen in the wild if the json is fubar, and the developer is notified of them by whatever crash reporting service they use. They fix the json, and the exceptions go away.

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageMetadata.java
thecombatwombat: Removing this TODO, since I'm told localization is something we'll look at post 1.0 GA

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
thecombatwombat: How do you propose we do that, other than the logging in question here? I don't think there is any sense in throwing an exception since they can't catch it here, and the user can go remove the permission in the settings anytime they wanted to.

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
thecombatwombat: Filed `AND-1227` for improvements/fixes to this class

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
thecombatwombat: Filed `AND-1227` for improvements/fixes to this class

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
thecombatwombat: Filed `AND-1227` for improvements/fixes to this class

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
thecombatwombat: Filed `AND-1227` for improvements/fixes to this class

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private static class GoogleApiCallbacks implements GoogleApiClient.OnConnectionFailedListener, GoogleApiClient.ConnectionCallbacks {
+        @Override
+        public void onConnected(Bundle bundle) {
+            if (Log.isLoggable(Log.VERBOSE)) Log.v("GoogleApiClient connected");
+        }
+
+        @Override
+        public void onConnectionSuspended(int i) {
+            if (Log.isLoggable(Log.VERBOSE)) Log.v("GoogleApiClient suspended");
```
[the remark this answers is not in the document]
thecombatwombat: Filed `AND-1227` for improvements/fixes to this class

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        @Override
+        public void onConnectionFailed(ConnectionResult connectionResult) {
+            if (Log.isLoggable(Log.VERBOSE)) Log.v("GoogleApiClient failed: " + connectionResult);
```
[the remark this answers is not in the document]
thecombatwombat: Filed `AND-1227` for improvements/fixes to this class

thecombatwombat:
Filed `AND-1227` for improvements/fixes to this class

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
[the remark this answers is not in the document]
smpete: Why should they be discoverable? It seems like they'd only be used when subclassing, no?

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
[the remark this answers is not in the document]
smpete: Usually a no-op as a result of a user action (i.e. editing the message) is poor UX. You don't know the result of your action.

I don't think a hard coded toast is a good idea. We should likely surface this to the developer so they can handle with a toast or dialog. Do we have an open task for that?

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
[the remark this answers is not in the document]
smpete: I don't see how that necessitates public visibility. In order to change this value you have to subclass.

You do bring up a good point about request code conflicts though. I'm not sure how we should handle that but we can move that to a separate task since that's a more widespread problem

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
smpete: Likely by surfacing that error to an error handler they are listening to. We do need to think about error propagation eventually. I'm good with leaving this as is for now

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
smpete: Why does it need to be discoverable? Expanding variable (or constant in this case) scope just so the developer is aware of it doesn't seem like the best solution in my opinion. Talk me out of that.

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
smpete: My question then is, is recovery possible for invalid JSON here or should it crash the app?

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
smpete: Right. Carry on

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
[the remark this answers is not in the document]
thecombatwombat: We have an open task for improving this class, an we can do ☝️ as part of that. But fundamentally, if the user does not have an app to handle that intent, not much us or the developer _can_ do. No-op. 

But I'm open to revisiting

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static final String ACTION_EVENT = "open-map";
+
+    public static final String KEY_ADDRESS = "address";
+    public static final String KEY_LATITUDE = "latitude";
+    public static final String KEY_LONGITUDE = "longitude";
+    public static final String KEY_ZOOM = "zoom";
+    public static final String KEY_TITLE = "title";
+
+    public static final int DEFAULT_ZOOM = 17;
```
thecombatwombat: make private

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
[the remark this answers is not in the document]
thecombatwombat: Keep keys public.

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import android.support.annotation.Nullable;
+
+import com.google.gson.JsonObject;
+import com.layer.sdk.LayerClient;
+
+import java.util.Locale;
+
+public class GoogleMapsOpenMapActionHandler extends ActionHandler {
+
+    public static final String ACTION_EVENT = "open-map";
```
thecombatwombat: make private

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: put ACTION_EVENT_OPEN_MAP in a constants file

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: IMO no. The message should be edited to have correct json.

layer-atlas/src/main/java/com/layer/ui/message/action/GoogleMapsOpenMapActionHandler.java
thecombatwombat: I'm going to file a separate ticket for putting all these strings into a separate constants class, since this will affect other actions as well.

layer-atlas/src/main/java/com/layer/ui/message/location/CurrentLocationSender.java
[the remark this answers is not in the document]
thecombatwombat: Filed `AND-1228` to fix this.

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: Filed a ticket `AND-1228` to fix this

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
thecombatwombat: Fixed to return `null`

layer-atlas/src/main/java/com/layer/ui/message/location/LocationMessageModel.java
```diff
@@ -0,0 +1,176 @@
...
+    private final Gson mGson;
+
+    private LocationMessageMetadata mMetadata;
+
+    public LocationMessageModel(Context context, LayerClient layerClient) {
+        super(context, layerClient);
+        GsonBuilder gsonBuilder = new GsonBuilder();
+        gsonBuilder.setFieldNamingStrategy(new AndroidFieldNamingStrategy());
+        mGson = gsonBuilder.create();
+
```
smpete: You added unnecessary blank lines here

PR #301, opened 2017-12-20
Task/and 1010 link primitive

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/message/link/LinkMessageMetadata.java
```diff
@@ -0,0 +1,87 @@
...
+    public String getImageUrl() {
+        return mImageUrl;
+    }
+
+    @Nullable
+    public String getUrl() {
+        return mUrl;
+    }
+
+    public Action getAction() {
```
smpete: Does this need a `@NonNull` annotation?

layer-atlas/src/main/java/com/layer/ui/message/link/LinkMessageModel.java
```diff
@@ -0,0 +1,143 @@
...
+        return mLinkMessageMetadata != null ? mLinkMessageMetadata.getAuthor() : null;
+    }
+
+    @Override
+    public String getActionEvent() {
+        if (mLinkMessageMetadata != null && mLinkMessageMetadata.getAction() != null) {
+            return mLinkMessageMetadata.getAction().getEvent();
+        }
+
+        return "open-url";
```
smpete: This string should be a constant somewhere

layer-atlas/src/main/java/com/layer/ui/message/link/LinkMessageModel.java
```diff
@@ -0,0 +1,143 @@
...
+        return actionData;
+    }
+
+    @Override
+    public int getBackgroundColor() {
+        return isMessageFromMe() ? R.color.ui_text_message_view_background_me : R.color.ui_text_message_view_background_them;
+    }
+
+    @Override
+    public boolean getHasContent() {
```
smpete: We should probably rename this to just `hasContent()` but that can be done in a separate task

layer-atlas/src/main/java/com/layer/ui/message/link/LinkMessageMetadata.java
thecombatwombat: Actually needs a `@Nullable` . Thanks!

layer-atlas/src/main/java/com/layer/ui/message/link/LinkMessageModel.java
thecombatwombat: I would much prefer that naming, but databinding is a pain when it comes to naming, and it the only way it will behave is if you name it `getHasContent` rather than `hasContent` :-|

PR #314, opened 2018-01-23
Task/and 1003 carousel composite

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/message/MessagePartUtils.java
```diff
@@ -135,6 +138,26 @@ public static boolean isParentMessagePart(@NonNull MessagePart rootMessagePart,
...
+                @Override
+                public int compare(MessagePart messagePart, MessagePart otherPart) {
+                    Map<String, String> messagePartArguments = getMimeTypeArguments(messagePart);
+                    Map<String, String> otherPartArguments = getMimeTypeArguments(otherPart);
+
+                    String messagePartOrder = messagePartArguments!=null ? messagePartArguments.get(PARAMETER_ITEM_ORDER) : null;
+                    String otherPartOrder = otherPartArguments!=null ? otherPartArguments.get(PARAMETER_ITEM_ORDER) : null;
+
+                    if (messagePartOrder == null || otherPartOrder == null) {
+                        return Integer.MAX_VALUE;
```
smpete: This breaks the contract. Specifically `sgn(compare(x, y)) == -sgn(compare(y, x))`. See the javadocs. Also, compare usually returns `-1`, `0` or `1` since the only thing that matters is that it is negative, positive, or 0.

layer-atlas/src/main/java/com/layer/ui/message/MessagePartUtils.java
```diff
@@ -135,6 +138,26 @@ public static boolean isParentMessagePart(@NonNull MessagePart rootMessagePart,
...
         }
 
+        if (!children.isEmpty()) {
+            Collections.sort(children, new Comparator<MessagePart>() {
+                @Override
+                public int compare(MessagePart messagePart, MessagePart otherPart) {
+                    Map<String, String> messagePartArguments = getMimeTypeArguments(messagePart);
+                    Map<String, String> otherPartArguments = getMimeTypeArguments(otherPart);
+
+                    String messagePartOrder = messagePartArguments!=null ? messagePartArguments.get(PARAMETER_ITEM_ORDER) : null;
```
smpete: Missing spaces before `!=`

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
```diff
@@ -199,10 +202,19 @@ public void onProgressError(MessagePart messagePart, Operation operation, Throwa
...
+        mAction = action;
+    }
+
+    @CallSuper
+    @Nullable
+    public String getActionEvent() {
+        return mAction != null ? mAction.getEvent() : null;
+    }
 
+    @CallSuper
```
smpete: Nice

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
```diff
@@ -199,10 +202,19 @@ public void onProgressError(MessagePart messagePart, Operation operation, Throwa
...
     public JsonObject getActionData() {
-        return new JsonObject();
+        return  mAction!=null ? mAction.getData() : new JsonObject();
```
smpete: We should also add a `@NonNull` annotation

layer-atlas/src/main/res/values-sw720dp/ui-dimens.xml
```diff
@@ -19,4 +19,5 @@
     <!-- # Recycler Maximum Cell Height Dimension                                             # -->
     <!-- ####################################################################################### -->
     <dimen name="layer_ui_messages_max_cell_height">876dp</dimen>
+    <dimen name="ui_text_message_view_max_width">280dp</dimen>
```
smpete: Is this for tablet layouts?

thecombatwombat: Yeah... and I guess a reminder to also optimize for landscape layouts

layer-atlas/src/main/java/com/layer/ui/message/MessagePartUtils.java
thecombatwombat: what conditions should we use then, given that `item-order` is optional and may be missing from some parts... 

I'm happy to change it over to `-1,0,1` but lets iron out the conditions

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
thecombatwombat: I can add that in

layer-atlas/src/main/res/values-sw720dp/ui-dimens.xml
smpete: Smallest width doesn't change in a landscape orientation. It's the literal smallest width of the device.

layer-atlas/src/main/java/com/layer/ui/message/MessagePartUtils.java
smpete: Well. `null` == `null` should return `0` obviously since those are equal. I assume we want null to always be less than an integer. So if message part order is non null and other part order is null we'd return 1. Else if other part order is non null and message part order is null then we'd return -1. Just following the same rules as https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#compare(int,%20int)

thecombatwombat: Well.. the `null`s I agree on... but for the `<1` condition.. I think it should be the other way around. 

If `messagePartOrder` is null and `otherPartOrder` is not, then it should receive priority and be "lesser" or higher up in the list of parts displayed right? First display the parts with an order, in order, and then everything else w/o an order

thecombatwombat: Just for posterity... we're doing this the other way around. 

`messagePartOrder < null` hence returns `-1` 

This puts the unordered parts at the end of the array, in random order.

layer-atlas/src/main/res/values-sw720dp/ui-dimens.xml
thecombatwombat: oh.. .removing now

PR #315, opened 2018-01-23
AND-1262 - Keeping local state of the choice buttons

smpete:
Fixing choice button set layout for choice messages. They denote a vertical layout with all choices in the set in it.
Fixing typo in choice metadata
Correctly handling button states for metadata flag
Notifying status view model of a change
Sending comma separated response representing full choice set

### DIFF
--- a/layer-atlas/src/main/java/com/layer/ui/message/choice/ChoiceButtonSet.java
+++ b/layer-atlas/src/main/java/com/layer/ui/message/choice/ChoiceButtonSet.java
@@ -17,18 +17,19 @@ import com.layer.ui.R;
 import com.layer.ui.util.Log;
 
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public class ChoiceButtonSet extends LinearLayout {
     private ColorStateList mChoiceButtonColorStateList;
     private boolean mAllowReselect;
     private boolean mAllowDeselect;
     private boolean mAllowMultiSelect;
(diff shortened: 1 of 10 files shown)


layer-atlas/src/main/java/com/layer/ui/message/choice/ChoiceButtonSet.java
```diff
@@ -43,7 +44,7 @@ public ChoiceButtonSet(Context context, @Nullable AttributeSet attrs, int defSty
         setLayoutParams(new ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT,
                 ViewGroup.LayoutParams.WRAP_CONTENT));
 
-        setOrientation(HORIZONTAL);
+        setOrientation(VERTICAL);
```
thecombatwombat: 🤔 How will this work in the context of a choice in the Buttons message?

smpete: I figured that would be done in the buttons work. Shouldn't be hard to bubble the horizontal vs vertical switch up to the encapsulating parent. I can add that here if you want.

thecombatwombat: That'd be a good idea... since afaik orientation needs to be set before subviews are added... but If I'm wrong or thats not needed, I can do it in buttons

PR #316, opened 2018-01-23
Feature/remove swipe to delete

thecombatwombat:


layer-atlas/src/main/res/layout/ui_four_part_item.xml
```diff
@@ -33,47 +34,24 @@
             type="FourPartItemViewModel"/>
     </data>
 
-    <com.layer.ui.util.views.SwipeableItem
+    <FrameLayout
```
thecombatwombat: @smpete this wonky little `FrameLayout` is needed since `ConstraintLayout` seems to not understand the meaning of `match_parent` when it itself is the root view ¯\_(ツ)_/¯ 

I've put it in there now, we can always find efficiencies for this later

PR #317, opened 2018-01-24
And 1260/fix typing indicator crash

andela-wakinsanmi:
Fix typing indicator crash


layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
```diff
@@ -371,7 +371,7 @@ public int getItemCount() {
 
     @Override
     public int getItemViewType(int position) {
-        if ((mShouldShowHeader && mHeaderView != null && position == getHeaderPosition()) || getItem(position) == null) {
+        if ((mShouldShowHeader && position == getHeaderPosition() && mHeaderView != null) || (position == getHeaderPosition() && getItem(position) == null)) {
```
smpete: Can we simplify this check to 
`
position == getHeaderPosition() && (mHeaderView != null || getItem(position) != null)
`

Does that make sense?

layer-atlas/src/main/res/values/ui-dimens.xml
```diff
@@ -63,6 +63,7 @@
     <dimen name="layer_ui_message_item_cell_padding_bottom">8dp</dimen>
     <dimen name="layer_ui_message_item_cell_padding_near">12dp</dimen>
     <dimen name="layer_ui_message_item_cell_padding_far">12dp</dimen>
+    <dimen name="layer_ui_message_item_footer_padding">30dp</dimen>
```
thecombatwombat: This attribute is used for a margin, so please name it `ui_message_item_footer_margin`

layer-atlas/src/main/res/layout/ui_message_item_footer.xml
```diff
@@ -28,6 +28,9 @@
     <RelativeLayout
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:layout_marginStart="@dimen/layer_ui_message_item_footer_padding"
```
smpete: Why does the footer only have start/left padding? Do we need padding on both sides?

layer-atlas/src/main/java/com/layer/ui/message/MessagesAdapter.java
thecombatwombat: Could this condition not be written as 
```
if ( position == getHeaderPosition() && ((mShouldShowHeader && mHeaderView != null) || (getItem(position) == null))) {```

PR #318, opened 2018-01-25
Task/and 1231 product message

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/message/product/ProductMessageView.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public ProductMessageView(Context context, @Nullable AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+        mBinding = UiProductMessageBinding.inflate(LayoutInflater.from(context), this, true);
+    }
+
+    @Override
+    public void setMessageModel(ProductMessageModel model) {
+        mModel = model;
```
smpete: We don't need this variable declaration here. The model parameter can be final and be referenced by the click listener. Or you can call `mBinding.getViewModel()`.

layer-atlas/src/main/java/com/layer/ui/message/product/ProductOptionsView.java
```diff
@@ -0,0 +1,71 @@
...
+        if (mOptions != null && !mOptions.isEmpty()) {
+            for (ChoiceMessageModel option : mOptions) {
+                String choiceId = option.getSelectedChoices() != null ? option.getSelectedChoices().get(0) : null;
+                List<ChoiceMetadata> choices = option.getChoiceMessageMetadata() != null ? option.getChoiceMessageMetadata().getChoices() : null;
+
+                if (choices != null && choices.size() > 0) {
+                    for (ChoiceMetadata choice : choices) {
+                        if (choice.getId().equals(choiceId)) {
+                            // Instantiate and add view
+                            UiProductOptionViewItemBinding binding = UiProductOptionViewItemBinding.inflate(LayoutInflater.from(getContext()), this, true);
```
smpete: Don't these choices need click listeners?

thecombatwombat: Not right now

PR #322, opened 2018-01-25
Have message models provide preview texts

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/conversationitem/ConversationItemFormatter.java
```diff
@@ -84,16 +95,40 @@ public String getTimeStamp(Conversation conversation) {
 
     public String getLastMessagePreview(Conversation conversation) {
         Message message = conversation.getLastMessage();
+        if (message == null) return "";
+
+        if (!mBinderRegistry.isLegacyMessageType(message)) {
+            String previewText = null;
+            String modelIdentifier = MessagePartUtils.getRootMimeType(message);
+            if (modelIdentifier != null && mBinderRegistry.getMessageModelManager().hasModel(modelIdentifier)) {
+                MessageModel model = mBinderRegistry.getMessageModelManager().getNewModel(modelIdentifier);
```
smpete: Yeah it's painful that we have to create a new model every time, especially since this is done at bind time. I'm okay with this for now as it's kind of the only way to do it. We should file an improvement task though so we don't forget about it

layer-atlas/src/main/java/com/layer/ui/conversationitem/ConversationItemFormatter.java
```diff
@@ -84,16 +95,40 @@ public String getTimeStamp(Conversation conversation) {
...
+            }
 
-        if (message!= null && mCellFactories != null && !mCellFactories.isEmpty()) {
-            for (CellFactory cellFactory : mCellFactories) {
-                if (cellFactory.isType(message)) {
-                    return cellFactory.getPreviewText(mContext, message);
+            previewText = previewText != null ? previewText : mContext.getString(R.string.ui_generic_message_preview_text);
+            boolean senderIsMe = message.getSender().equals(mLayerClient.getAuthenticatedUser());
+            if (senderIsMe) {
+                return String.format("%s %s", mContext.getString(R.string.ui_preview_text_you_sent),
```
smpete: Formatting should be in the xml file. Use `getString(int resId, Object... formatArgs)`

layer-atlas/src/main/java/com/layer/ui/conversationitem/ConversationItemFormatter.java
```diff
@@ -84,16 +95,40 @@ public String getTimeStamp(Conversation conversation) {
...
+                        previewText);
+            } else {
+                return String.format("%s %s %s", mIdentityFormatter.getDisplayName(message.getSender()),
```
smpete: Formatting should be in the xml file. Use getString(int resId, Object... formatArgs)

layer-atlas/src/main/java/com/layer/ui/message/text/TextMessageModel.java
```diff
@@ -106,6 +107,16 @@ public boolean getHasContent() {
         return !TextUtils.isEmpty(mText);
     }
 
+    @Nullable
+    @Override
+    public String getPreviewText() {
+        if (getHasContent()) {
+            return mText.length() > 100 ? mText.substring(0, 100) : mText;
```
smpete: Define 100 as a constant somewhere

layer-atlas/src/main/java/com/layer/ui/conversationitem/ConversationItemFormatter.java
thecombatwombat: If we're happy with the previews coming in from only top level/root models, I wonder if we could write an interface/abstract class that is optionally registered alongside the registry call in `MessageModelManager`

Something like 
```interface MessagePreviewFormatter<T extends MessageModel> {
     public String getPreview(Message message);
}```

Comment (thecombatwombat) [reply layer-atlas/src/main/java/com/layer/ui/conversationitem/ConversationItemFormatter.java]: Filed `AND-1282`

File: layer-atlas/src/main/res/values/ui-strings.xml [hunk 5]
```diff
@@ -88,4 +88,30 @@
...
+    </plurals>
+
+    <string name="ui_product_message_preview_text">a product</string>
+    <string name="ui_file_message_preview_text">a file</string>
+    <string name="ui_image_message_preview_text">an image</string>
+    <string name="ui_location_message_preview_text">a location</string>
+    <string name="ui_link_message_preview_text">a link</string>
+    <string name="ui_text_message_preview_text">a message</string>
+    <string name="ui_generic_message_preview_text">a message</string>
+    <string name="ui_preview_text_you_sent">You:  %s</string>
```
smpete: Why do these have two spaces between the semicolon and the preview text?

thecombatwombat: The correct answer to that question is -> I'm stupid. 

Fixing now

PR #324, opened 2018-01-25
fix tint issue on Lollipop

andela-wakinsanmi:
Add check for pre lollipop to fix the tint issue

### DIFF
--- a/layer-atlas/src/main/java/com/layer/ui/composebar/ComposeBar.java
+++ b/layer-atlas/src/main/java/com/layer/ui/composebar/ComposeBar.java
@@ -299,12 +299,9 @@ public class ComposeBar extends FrameLayout implements TextWatcher {
 
         if (sender.getIcon() != null) {
             Drawable iconDrawable = ContextCompat.getDrawable(getContext(), sender.getIcon());
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-                DrawableCompat.setTint(iconDrawable, getResources().getColor(R.color.layer_ui_icon_enabled));
-            } else {
-                iconDrawable.mutate().setColorFilter(getResources().getColor(R.color.layer_ui_icon_enabled), PorterDuff.Mode.SRC_IN);
-            }
-            binding.title.setCompoundDrawablesWithIntrinsicBounds(iconDrawable, null, null, null);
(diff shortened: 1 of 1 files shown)


layer-atlas/src/main/java/com/layer/ui/composebar/ComposeBar.java
```diff
@@ -298,7 +299,11 @@ protected void addAttachmentMenuItem(AttachmentSender sender) {
 
         if (sender.getIcon() != null) {
             Drawable iconDrawable = ContextCompat.getDrawable(getContext(), sender.getIcon());
-            DrawableCompat.setTint(iconDrawable, getResources().getColor(R.color.layer_ui_icon_enabled));
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                DrawableCompat.setTint(iconDrawable, getResources().getColor(R.color.layer_ui_icon_enabled));
```
smpete: Shouldn't DrawableCompat negate the need to do this check? What is wrong with using DrawableCompat?

andela-wakinsanmi: Yes, the problem with that is the setTint doesn't work well with the Drawable on Lollipop. Another approach will be to use a wrapped drawable. 
       Drawable wrappedDrawable = DrawableCompat.wrap(iconDrawable); I

smpete: Why doesn't it work? Do you just need to set the tint mode on the drawable? This is kind of the point of `DrawableCompat` so there is likely an issue elsewhere. Unless you can explain why `DrawableCompat` will not work let's keep investigating this.

andela-wakinsanmi: I mean ```  if (sender.getIcon() != null) {
            Drawable iconDrawable = ContextCompat.getDrawable(getContext(), sender.getIcon());
            Drawable wrappedDrawable = DrawableCompat.wrap(iconDrawable);
            DrawableCompat.setTint(wrappedDrawable, getResources().getColor(R.color.layer_ui_icon_enabled));
            binding.title.setCompoundDrawablesWithIntrinsicBounds(wrappedDrawable, null, null, null);
        }```

PR #325, opened 2018-01-25
Fix drawable for file

andela-wakinsanmi:


smpete:
Can you explain this change? That's an incorrect asset for the attachment menu

andela-wakinsanmi:
The asset we had was 70 X 80px, that's the reason it show too big. What I have done is to use Preview to resize it to appropriate size.

smpete:
That asset looks incorrect anyways. Perhaps this will be fixed when `AND-1266` is fixed? Those added drawables are pixelated as well. Where did you get them?

andela-wakinsanmi:
From the original one in drawable-xhdpi

smpete:
We can't just scale up drawables to a larger size

layer-atlas/src/main/res/layout/ui_file_message_layout.xml
```diff
@@ -23,7 +23,7 @@
             android:layout_height="@dimen/layer_ui_file_message_image_height"
             android:adjustViewBounds="true"
             android:scaleType="center"
-            android:src="@drawable/ui_ic_file_generic"
+            android:src="@drawable/ic_insert_drive_file_white_24dp"
```
smpete: This shouldn't change. This is used for the message preview, not the attachment menu.

layer-atlas/src/main/java/com/layer/ui/message/file/FileMessageModel.java
```diff
@@ -167,7 +167,7 @@ public int getFileIconDrawable() {
     }
 
     private void setupFileIconDrawable(@Nullable String sourceMimeType) {
-        mFileIconDrawable = R.drawable.ui_ic_file_generic;
+        mFileIconDrawable = R.drawable.ic_insert_drive_file_white_24dp;
```
smpete: Nor should this change. This is not in the attachment sender menu

PR #327, opened 2018-01-25
AND-1280 - Using correct parent node id for image and file sending

smpete:


thecombatwombat:
@smpete LGTM... but I think the same might be needed for Location messages, esp for when they're nested in some composite

smpete:
We don't need to handle that right now since we're not composing a location message as a child node yet.

PR #329, opened 2018-01-30
Task/and 1002 receipt message

thecombatwombat:


layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
```diff
@@ -239,6 +245,37 @@ public boolean getHasMetadata() {
...
+    public String getRole() {
+        return mRole;
+    }
+
+    public void setRole(@NonNull String role) {
+        mRole = role;
+    }
+
+    @NonNull
+    public List<MessageModel> getChildMessageModelsWithRole(@NonNull String role) {
```
smpete: Should this be a `Set` instead of a `List`? Actually maybe we should change those container member variables to sets instead of lists if we're not ordering those. That should likely be done in a separate task.

layer-atlas/src/main/java/com/layer/ui/message/product/ProductMessageModel.java
```diff
@@ -208,4 +216,29 @@ public ImageRequestParameters getImageRequestParameters() {
...
         return mOptions;
     }
+
+    @Nullable
+    public String getSelectedOptionsAsCommaSeparatedList() {
+        List<String> productTexts = new ArrayList<>();
+        List<ChoiceMessageModel> options = getOptions();
+        if (options != null && !options.isEmpty()) {
+            for (ChoiceMessageModel option : options) {
+                Iterator<String> iterator = option.getSelectedChoices() != null ? option.getSelectedChoices().iterator() : null;
```
smpete: Are we not supposed to iterate through these selected choice ids?

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public Class<? extends MessageView> getRendererType() {
+        return ReceiptMessageView.class;
+    }
+
+    @Override
+    protected void parse(@NonNull MessagePart messagePart) {
+        if (getRootMessagePart() == messagePart) {
```
smpete: Use object equality instead of reference equality

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
```diff
@@ -0,0 +1,133 @@
...
+            JsonReader reader = new JsonReader(new InputStreamReader(messagePart.getDataStream()));
+            mMetadata = mGson.fromJson(reader, ReceiptMessageMetadata.class);
+        }
+    }
+
+    @Override
+    protected boolean shouldDownloadContentIfNotReady(@NonNull MessagePart messagePart) {
+        return false;
```
smpete: Why is this false by default?

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public boolean getHasContent() {
+        return mMetadata != null;
+    }
+
+    @Nullable
+    @Override
+    public String getPreviewText() {
+        return null;
```
smpete: Shouldn't this have preview text?

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageProductItemView.java
```diff
@@ -0,0 +1,48 @@
...
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.widget.FrameLayout;
+
+import com.layer.ui.R;
+import com.layer.ui.databinding.UiReceiptMessageProductDetailBinding;
+import com.layer.ui.message.product.ProductMessageModel;
+
+public class ReceiptMessageProductItemView extends FrameLayout {
+    private LayoutInflater mInflater;
```
smpete: This can be a local variable

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageProductItemView.java
```diff
@@ -0,0 +1,48 @@
...
+        super(context, attrs, defStyleAttr);
+        mInflater = LayoutInflater.from(context);
+        mBinding = UiReceiptMessageProductDetailBinding.inflate(mInflater, this, true);
+    }
+
+    public void setProductModel(ProductMessageModel productModel) {
+        mBinding.setViewModel(productModel);
+        String subtitle = productModel.getSelectedOptionsAsCommaSeparatedList();
+        if (subtitle != null) {
+            String text = getContext().getResources().getString(R.string.ui_receipt_product_detail_options, subtitle);
```
smpete: `getResources()` can be omitted.

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageProductItemView.java
```diff
@@ -0,0 +1,48 @@
...
+            mBinding.subtitle.setText(text);
+        } else if (productModel.getMetadata() != null) {
+            String quantity = productModel.getMetadata().getQuantity() > 0 ?
+                    getContext().getResources().getString(R.string.ui_receipt_product_detail_quantity,
```
smpete: `getResources()` can be omitted.

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageView.java
```diff
@@ -0,0 +1,90 @@
...
+import com.layer.ui.message.container.TitledMessageContainer;
+import com.layer.ui.message.location.LocationMessageModel;
+import com.layer.ui.message.product.ProductMessageModel;
+import com.layer.ui.message.view.MessageView;
+
+import java.util.List;
+
+public class ReceiptMessageView extends MessageView<ReceiptMessageModel> {
+    private UiReceiptMessageBinding mBinding;
+    private LayoutInflater mInflater;
```
smpete: This can also be a local variable

layer-atlas/src/main/res/layout/ui_receipt_message.xml
```diff
@@ -0,0 +1,176 @@
...
+                android:textColor="@color/ui_receipt_message_view_detail_value_text"
+                android:textSize="@dimen/ui_receipt_message_view_detail_value_size"
+                app:visibleOrGone="@{viewModel.billingAddressLocationModel!=null &amp;&amp; viewModel.billingAddressLocationModel.metadata!=null}"/>
+
+        </LinearLayout>
+
+        <android.support.constraint.ConstraintLayout
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            android:paddingBottom="@dimen/layer_ui_margin_large"
```
smpete: Replace all of these paddings with `android:padding`

layer-atlas/src/main/java/com/layer/ui/message/model/MessageModel.java
thecombatwombat: Potentially....

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
thecombatwombat: Why would the outcome of that be different?

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
thecombatwombat: Should be true. Fixing

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
thecombatwombat: It should... fixing

layer-atlas/src/main/java/com/layer/ui/message/receipt/ReceiptMessageModel.java
smpete: The outcome will not be different currently since each `MessagePart` has one instance in the SDK. With that said, the other `MessageModel`s are using object equality so I think we should continue that paradigm.

In regards to this public API changing, let's create another task and discuss that there since it's not specific to this receipt work.

layer-atlas/src/main/java/com/layer/ui/message/product/ProductMessageModel.java
thecombatwombat: Afaik... those will be useful in the full, expanded product card, which is to be worked on at a later date

